PR #94, opened 2025-12-02
[EX-1234] Add flagevaluation tracking event emission.

lisaah:
## Motivation

We want to switch to using the flagevaluation track to keep track of feature flag evaluations. This is an example of implementing for the browser.

Mainly looking for thoughts on approach on having the aggregator for collecting the events.

~~Note: when the freeze is over, will want to bump / release the version of browser sdk and use that.~~

## Test instructions

I've been building and running locally and checking the events are being propagated into the backend on staging.

image

## Checklist

- [ ] Updated Documentation
- [ ] Tested locally
- [ ] Tested on staging

### DIFF
--- a/packages/browser/test/openfeature/flagEvaluations.spec.ts
+++ b/packages/browser/test/openfeature/flagEvaluations.spec.ts
@@ -37,7 +37,6 @@ jest.mock('@datadog/browser-core', () => ({
   dateNow: jest.fn(() => 1234567890),
 }))
 
-
 describe('createFlagEvaluationTrackingHook', () => {
   it('should create a hook that tracks flag evaluations', () => {
     const hook = createFlagEvaluationTrackingHook(mockConfiguration)


packages/browser/src/domain/configuration.ts
```diff
@@ -46,6 +47,16 @@ export interface FlaggingInitConfiguration extends InitConfiguration {
...
 
+  /**
+   * Whether to enable flag evaluation tracking via the flag evaluation intake
+   */
+  enableFlagEvaluationTracking?: boolean
+
+  /**
+   * Flag evaluation tracking interval in milliseconds (default: 10000ms)
+   */
+  flagEvaluationTrackingInterval?: number
```
petzel: Where is the default for this set?

edit: i think it is from `DEFAULT_FLAG_EVALUATION_TRACKING_INTERVAL`. IMO it would be cleaner to have the default set here in this object rather than the downstream. having the default only set in a comment can lead to it getting out of sync with the actual default, and it makes changing the default more difficult as one needs to trace through all of the code.

packages/browser/src/transport/startFlagEvaluationBatch.ts
```diff
@@ -0,0 +1,34 @@
...
+import {
+  createBatch,
+  createFlushController,
+  createHttpRequest,
+  createIdentityEncoder,
+  Observable,
+} from '@datadog/browser-core'
+import type { FlaggingConfiguration } from '../domain/configuration'
+
+export function startFlagEvaluationBatch(
```
petzel: Why do we need this additional method + file? Could we put this logic directly in `createBatch`?

packages/browser/src/domain/configuration.ts
```diff
@@ -82,6 +97,9 @@ export function validateAndBuildFlaggingConfiguration(
 
   return {
     applicationId: initConfiguration.applicationId,
+    flagEvaluationTrackingInterval: initConfiguration.flagEvaluationTrackingInterval,
+    // [FlagEval] TODO: Don't set this once we have a proper endpoint builder from browser core SDK
```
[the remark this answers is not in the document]
lisaah: This needs to be released https://github.com/DataDog/browser-sdk

PR #141, opened 2026-01-13
fix(core): include empty string targeting keys in flag evaluation events

typotter:
## Motivation

The truthiness check for `targetingKey` in flag evaluation events was incorrectly excluding empty strings, which are valid targeting keys. This meant that evaluations with empty string targeting keys would omit the `targeting_key` field entirely from the event payload.

## Changes
 use `if (data.targetingKey != null)` instead of `if (data.targetingKey)`

## Test instructions

Run the test suite for the new test file:
```bash
cd packages/core && yarn test flagEvaluationEvent.spec
```

All 7 tests should pass, specifically verifying that:
- `targeting_key` is included when it's an empty string
- `targeting_key` is not included when it's undefined

## Checklist

- [ ] Updated Documentation
- [x] Tested locally
- [ ] Tested on staging

### DIFF
--- a/packages/core/src/configuration/flagEvaluationEvent.ts
+++ b/packages/core/src/configuration/flagEvaluationEvent.ts
@@ -27,7 +27,9 @@ export function createFlagEvaluationEvent(data: FlagEvaluationAggregationData, t
     timestamp,
   }
 
-  event.targeting_key = data.targetingKey ?? ''
+  if (data.targetingKey != null) {
+    event.targeting_key = data.targetingKey
+  }
 
   if (data.error) {
     event.error = { message: data.error }
--- a/packages/core/test/configuration/flagEvaluationEvent.spec.ts
+++ b/packages/core/test/configuration/flagEvaluationEvent.spec.ts
@@ -34,7 +34,7 @@ describe('createFlagEvaluationEvent', () => {
     expect(event).toHaveProperty('targeting_key')
   })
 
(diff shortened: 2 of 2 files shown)


packages/core/src/configuration/flagEvaluationEvent.ts
```diff
@@ -27,7 +27,7 @@ export function createFlagEvaluationEvent(data: FlagEvaluationAggregationData, t
     timestamp,
   }
 
-  if (data.targetingKey) {
+  if (data.targetingKey != null) {
     event.targeting_key = data.targetingKey
   }
```
dd-oleksii: 🙂 
> **EVAL.7** If targeting key is missing (undefined) from the evaluation context, the evaluation event must report be sent with targeting_key: "".

```suggestion
  event.targeting_key = data.targetingKey ?? ''
```

packages/core/test/configuration/flagEvaluationEvent.spec.ts
```diff
@@ -0,0 +1,141 @@
...
+        runtimeDefaultUsed: false,
+      },
+      2000
+    )
+
+    expect(event.targeting_key).toBe('')
+    expect(event).toHaveProperty('targeting_key')
+  })
+
+  it('should not include targeting_key when it is undefined', () => {
```
dd-oleksii: See EVAL.7

PR #186, opened 2026-02-26
Add IndexedDB persistence for browser flag configurations

leoromanovsky:
## Motivation

The browser SDK currently has no disk persistence for flag configurations. If the network is unavailable during a page load (backend outage, flaky connection, CDN maintenance), users get no flags and fall back to programmatic defaults. iOS and Android SDKs both have disk-based caching that makes them resilient to short outages — this brings the browser SDK to parity.

## Changes

The provider now stores the full precomputed flag configuration as a JSON blob in IndexedDB after every successful fetch. On the next page load, it reads the cached configuration from IndexedDB before hitting the network. If the network request fails but cached flags exist, the provider goes to READY state using the cached data instead of throwing an error. If IndexedDB is unavailable (private browsing, older browsers), the SDK works exactly as it does today — no behavior change.

## Decisions

- **IndexedDB over localStorage** — localStorage is synchronous and blocks the main thread on read/write. It also has a ~5MB limit shared across all keys for the origin, and we'd be competing with the exposure assignment cache that already uses localStorage. IndexedDB is async, has much higher storage limits (~50MB+), and keeps flag persistence isolated from other SDK storage.
- **Single JSON blob, not individual flags in IndexedDB** — Simpler, matches mobile SDK patterns, and the payload is small (a few KB). No need for indexed queries on individual flags.
- **Direct `JSON.stringify`/`JSON.parse` instead of the wire format** — The existing `configurationToString`/`configurationFromString` has a round-trip bug where it nests the response one level deep on deserialization. Direct serialization avoids this.
- **Network-first, cache-fallback** — Always attempt a fresh fetch. IndexedDB is only used as a fallback during `initialize()` when the network fails.
- **No TTL** — Matches iOS and Android. Cached data is used until a successful network fetch replaces it.
- **Graceful degradation** — If IndexedDB is unavailable, `flagsCache` is undefined and all optional-chain calls are no-ops.

## Question for reviewers

This is a non-breaking additive feature (no API changes, no behavior change when IndexedDB is unavailable). Should this ship as a **patch** release or a **minor** release?

### DIFF
--- a/packages/browser/src/openfeature/provider.ts
+++ b/packages/browser/src/openfeature/provider.ts
@@ -100,9 +100,7 @@ export class DatadogProvider implements Provider {
     }
 
     // Start all async work concurrently — cache read should not delay the fetch.
-    const cachedConfigPromise = !this.hasInitialFlagsConfiguration
-      ? this.flagsCache?.get(context)
-      : undefined
+    const cachedConfigPromise = !this.hasInitialFlagsConfiguration ? this.flagsCache?.get(context) : undefined
     const exposureCacheReady = this.exposureCache?.init()
 
     try {


packages/browser/src/openfeature/provider.ts
```diff
@@ -47,6 +48,7 @@ export class DatadogProvider implements Provider {
   private flagsConfiguration: FlagsConfiguration = {}
   private configuration?: FlaggingConfiguration
   private exposureCache: AssignmentCache | undefined
+  private flagsCache: IndexedDBFlagsCache | undefined
```
leoromanovsky: I'll probably refactor this into an abstract type

greghuels (approved):

packages/browser/src/cache/indexeddb-flags-cache.ts
```diff
@@ -0,0 +1,93 @@
...
+    request.onsuccess = () => resolve(request.result)
+    request.onerror = () => reject(request.error)
+  })
+}
+
+export class IndexedDBFlagsCache {
+  private readonly configKey: string
+
+  constructor(clientToken: string) {
+    this.configKey = `flags-config-${buildStorageKeySuffix(clientToken)}`
```
dd-oleksii: major: this may still conflict if there are multiple providers with different context. We should include targeting key and attributes into the key.

Will also need some way to evict old entries

PR #150, opened 2026-01-26
fix(core): resolve peer dependency errors for dd-trace-js users

leoromanovsky:
## Motivation

Users of `dd-trace` who don't directly use OpenFeature are encountering peer dependency errors when installing `dd-trace@5.75.0+`. This is tracked in https://github.com/DataDog/dd-trace-js/issues/6986

The issue occurs because `dd-trace-js` lists `@datadog/openfeature-node-server` as a **hard dependency**, which creates this transitive chain:

```
dd-trace
└── @datadog/openfeature-node-server (hard dependency)
    ├── @openfeature/server-sdk (was hard dependency)
    │   └── @openfeature/core (peer dependency - expects end user to install)
    └── @datadog/flagging-core
        └── @openfeature/core (peer dependency - expects end user to install)
```

Since `@openfeature/core` is a peer dependency, npm/yarn expects the **end user** to install it. But dd-trace users didn't ask for OpenFeature integration and don't know they need to install `@openfeature/core`, resulting in peer dependency errors and runtime failures.

## Reproducing the Issue

**package.json:**
```json
{
  "name": "repro-issue",
  "version": "1.0.0",
  "dependencies": {
    "dd-trace": "5.75.0"
  }
}
```

**test.js:**
```javascript
// Scenario 1: Just init dd-trace (most users)
console.log("=== Test 1: Basic dd-trace init ===");
try {
  require('dd-trace').init();
  console.log("✓ dd-trace init works\n");
} catch (e) {
  console.log("✗ dd-trace init failed:", e.message, "\n");
}

// Scenario 2: Try to use OpenFeature via dd-trace's bundled provider
console.log("=== Test 2: Use OpenFeature SDK ===");
try {
  const { OpenFeature } = require('@openfeature/server-sdk');
  console.log("✓ OpenFeature loaded\n");
} catch (e) {
  console.log("✗ OpenFeature failed:", e.message, "\n");
}

// Scenario 3: Try to use Datadog's OpenFeature provider
console.log("=== Test 3: Use Datadog OpenFeature provider ===");
try {
  const { DatadogProvider } = require('@datadog/openfeature-node-server');
  console.log("✓ DatadogProvider loaded\n");
} catch (e) {
  console.log("✗ DatadogProvider failed:", e.message, "\n");
}
```

**Before fix (with published packages):**
```
$ yarn add dd-trace@5.75.0
warning "@datadog/openfeature-node-server" has unmet peer dependency "@openfeature/server-sdk@~1.18.0"
[...]

PR #248, opened 2026-04-21
fix(node): mark open @openfeature/server-sdk as non-optional

dd-oleksii:
## Changes

We have a non-type import here: https://github.com/DataDog/openfeature-js-client/blob/fae56cc9b1c8484b798458dbc124bf15c1120908/packages/node-server/src/provider.ts#L21

So the package must be present for our provider to work.

## Checklist

- [ ] Updated Documentation
- [ ] Tested locally
- [ ] Tested on staging

### DIFF
--- a/packages/node-server/package.json
+++ b/packages/node-server/package.json
@@ -41,11 +41,6 @@
   "peerDependencies": {
     "@openfeature/server-sdk": ">=1.15.1"
   },
-  "peerDependenciesMeta": {
-    "@openfeature/server-sdk": {
-      "optional": true
-    }
-  },
   "devDependencies": {
     "@openfeature/core": "1.9.2",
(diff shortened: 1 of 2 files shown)


dd-oleksii:
@leoromanovsky do you remember why this metadata was needed? it looks incorrect

greghuels (approved):

dd-oleksii:
Looked a bit into the history and this was introduced in https://github.com/DataDog/openfeature-js-client/pull/150. Digging into original issue, it must have been fixed by removing `@openfeature/server-sdk` from `dependencies` list (only leaving it in `peerDependencies`), and marking it optional was an extraneous change.

dd-oleksii:
@BridgeAR you have suggested adding this optional bit in #150. Do you remember why?

PR #249, opened 2026-04-22
chore: ignore dependabot major version upgrades

greghuels:
## Motivation

Major version upgrades require more manual scrutiny to make sure that nothing is broken.

## Changes

Ignoring dependabot major version upgrades

### DIFF
--- a/.github/dependabot.yaml
+++ b/.github/dependabot.yaml
@@ -24,6 +24,10 @@ updates:
       prefix: "\U0001F477 chore(deps)"
       prefix-development: "\U0001F477 chore(deps-dev)"
       include: scope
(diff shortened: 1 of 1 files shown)


typotter (approved):

dd-oleksii (commented):
Dependabot is still helpful for major versions. In many cases one can review release notes (which dependabot helpfully pastes) and decide whether more scrutiny is needed

greghuels:
> Dependabot is still helpful for major versions. In many cases one can review release notes (which dependabot helpfully pastes) and decide whether more scrutiny is needed

Yeah there's definitely an argument to keep it. I can close this out.

PR #262, opened 2026-04-28
fix(browser): remove dead @datadog/browser-core/cjs/ subpath import

typotter:
## Summary

Removes a deep subpath import `@datadog/browser-core/cjs/domain/configuration` that was never in the package's exports map. TypeScript emits it verbatim into ESM output, causing bundler failures (Vite/Rollup/webpack) when strict subpath resolution is enabled — the root cause of the breakage reported in #247.

The `createEndpointBuilder` call it powered was always a dead assignment: the `...baseConfiguration` spread on the next line overwrites it. `validateAndBuildConfiguration` already returns `flagEvaluationEndpointBuilder` via `TransportConfiguration` (since browser-core 6.25.1).

**Stacks on:** #261 (browser-core bump to ^6.33.0 + transport API adaptation)

## Verification

```
PASS: CJS configuration output is clean
PASS: ESM configuration output is clean
PASS: flagEvaluationEndpointBuilder works correctly
      URL: https://browser-intake-datadoghq.com/api/v2/flagevaluation
PASS: trackType is "flagevaluation"
=== Results: 4 passed, 0 failed ===
```

## Test plan

- [ ] CI passes
- [ ] No `/cjs/` subpath imports remain in built output

### DIFF
--- a/packages/browser/src/domain/configuration.ts
+++ b/packages/browser/src/domain/configuration.ts
@@ -1,6 +1,5 @@
 import type { Configuration, EndpointBuilder, InitConfiguration } from '@datadog/browser-core'
 import { display, validateAndBuildConfiguration } from '@datadog/browser-core'
-import { createEndpointBuilder, type TrackType } from '@datadog/browser-core/cjs/domain/configuration'
 import type { FlagsConfiguration } from '@datadog/flagging-core'
 import type { EvaluationContext } from '@openfeature/web-sdk'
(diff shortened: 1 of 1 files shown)


typotter:
Superseded — configuration.ts fix moved into #261 (the TS2783 error proved the two changes are inseparable: the dead assignment now causes a type error with browser-core ^6.33.0 types).

PR #263, opened 2026-04-29
fix: avoid MapIterator in declarations to support TypeScript < 5.6

greghuels:
## Problem

When `@datadog/flagging-core` is built with TypeScript 5.6+, the emitted `.d.ts` files use `MapIterator` as the return type for iterator methods on classes that `implements Map<string, string>`. `MapIterator` was introduced in TypeScript 5.6 and does not exist in older versions, so any consumer on TypeScript < 5.6 (with `skipLibCheck: false`) gets:

```
error TS2304: Cannot find name 'MapIterator'
```

This was a regression introduced when the package was rebuilt with TS 5.6+.

## Fix

- Introduce a `CacheDelegate` interface with `IterableIterator<>` return types as a narrower replacement for the `Map<string, string>` constraint on `AbstractAssignmentCache<T>`
- Remove `implements Map<string, string>` from `LRUCache` and `LocalStorageAssignmentShim`; add explicit `IterableIterator<>` return type annotations on iterator methods so the emitted declarations are pinned to `IterableIterator` regardless of which TypeScript version builds the package
- Remove `Symbol.iterator`, `Symbol.toStringTag`, `forEach`, and `size` from `LRUCache` as these were only required to satisfy the `Map` interface

## Testing

- All 26 existing tests pass
- Both `@datadog/flagging-core` and `@datadog/openfeature-browser` build cleanly
- Confirmed no `MapIterator` references in any emitted `.d.ts` files

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/packages/core/src/cache/abstract-assignment-cache.ts
+++ b/packages/core/src/cache/abstract-assignment-cache.ts
@@ -41,7 +41,22 @@ export interface AssignmentCache {
   clear(): Promise<void> | void
 }
 
-export abstract class AbstractAssignmentCache<T extends Map<string, string>> implements AssignmentCache {
+/**
+ * Minimal interface for the backing store used by {@link AbstractAssignmentCache}.
+ *
+ * Using a narrower constraint instead of `Map<string, string>` avoids emitting
+ * `MapIterator` (introduced in TypeScript 5.6) in declaration files, which would
+ * break consumers on TypeScript < 5.6.
+ */
+export interface CacheDelegate {
+  get(key: string): string | undefined
+  set(key: string, value: string): this
+  has(key: string): boolean
+  clear(): void
+  entries(): IterableIterator<[string, string]>
+}
+
+export abstract class AbstractAssignmentCache<T extends CacheDelegate> implements AssignmentCache {
   // key -> variation value hash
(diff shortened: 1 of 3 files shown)


packages/core/src/cache/abstract-assignment-cache.ts
```diff
@@ -41,7 +41,22 @@ export interface AssignmentCache {
...
+/**
+ * Minimal interface for the backing store used by {@link AbstractAssignmentCache}.
+ *
+ * Using a narrower constraint instead of `Map<string, string>` avoids emitting
+ * `MapIterator` (introduced in TypeScript 5.6) in declaration files, which would
+ * break consumers on TypeScript < 5.6.
+ */
+export interface CacheDelegate {
+  get(key: string): string | undefined
+  set(key: string, value: string): this
```
typotter, line 53: the chaining with `this` here isn't needed as the return value by callers is ignored. I think `void` here makes more sense

typotter (commented):
just one non-blocker and maybe consider a name for `CacheDelegate` more specific like `AssignmentCacheStore` or `StringMapDelegate`

typotter (approved):

PR #269, opened 2026-05-05
feat(node): expose serialId in flagMetadata for span enrichment

sameerank:
## Motivation

Enable APM span enrichment by exposing `serialId` in the evaluation result's `flagMetadata`. This allows tracer SDKs to access the serial ID directly from the evaluation response and encode it into span tags (`ffe_flags_enc`).

Related tickets:
- [FFL-2221](https://datadoghq.atlassian.net/browse/FFL-2221) - Node.js implementation
- [FFL-2220](https://datadoghq.atlassian.net/browse/FFL-2220) - Parent ticket for all SDKs

## Changes

- Add `serialId?: number` to `Split` interface in `packages/node-server/src/configuration/ufc-v1.ts`
- Add `serialId?: number` to `PrecomputedFlagMetadata` type in `packages/core/src/configuration/configuration.ts`
- Pass `selectedSplit.serialId` in `flagMetadata` in `packages/node-server/src/configuration/evaluateForSubject.ts`

## Backwards Compatibility

The `serialId` field is optional (`serialId?: number`). This ensures:
- **UFC without serialId**: If the Universal Flag Configuration does not contain `serialId` on splits, `flagMetadata.serialId` will be `undefined`. Consuming code (tracers) must handle this gracefully.
- **No breaking changes**: Existing integrations continue to work without modification.

The dd-trace-js implementation handles missing `serialId` gracefully:
```javascript
const serialId = flagMetadata?.serialId
if (serialId != null) {
  // Only process if serialId exists
}
```

## Test instructions

1. Build the package: `yarn build`
2. Verify TypeScript compiles without errors
3. Integration testing will be done via system-tests parametric tests

## Checklist

- [ ] Updated Documentation
- [x] Tested locally
- [ ] Tested on staging

[FFL-2221]: https://datadoghq.atlassian.net/browse/FFL-2221?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[FFL-2220]: https://datadoghq.atlassian.net/browse/FFL-2220?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/core/src/configuration/configuration.ts
+++ b/packages/core/src/configuration/configuration.ts
@@ -53,18 +53,15 @@ export type PrecomputedFlag<T extends FlagValueType = FlagValueType> = {
 
 /** @internal */
 export type PrecomputedFlagMetadata = {
-  // Standardized keys
-  __dd_allocation_key: string
-  __dd_do_log: boolean
-  __dd_split_serial_id?: number
-
-  // Deprecated keys (will be removed in next major version)
-  /** @deprecated Use __dd_allocation_key instead */
+  // Primary keys (used by browser and core readers)
   allocationKey: string
-  /** @deprecated Not used externally, will be removed */
   variationType: FlagValueType
-  /** @deprecated Use __dd_do_log instead */
   doLog: boolean
+
+  // Server-side tracing keys (consumed by dd-trace-js)
+  __dd_allocation_key?: string
+  __dd_do_log?: boolean
+  __dd_split_serial_id?: number
 }
 
 /**
--- a/packages/core/src/configuration/configuration.ts
+++ b/packages/core/src/configuration/configuration.ts
@@ -56,6 +56,7 @@ export type PrecomputedFlagMetadata = {
   allocationKey: string
   variationType: FlagValueType
   doLog: boolean
+  serialId?: number
 }
 
 /**
--- a/packages/core/src/configuration/configuration.ts
+++ b/packages/core/src/configuration/configuration.ts
@@ -53,10 +53,18 @@ export type PrecomputedFlag<T extends FlagValueType = FlagValueType> = {
 
 /** @internal */
 export type PrecomputedFlagMetadata = {
-  allocationKey: string
-  variationType: FlagValueType
-  doLog: boolean
+  // Standardized keys
+  __dd_allocation_key: string
+  __dd_do_log: boolean
   __dd_split_serial_id?: number
+
+  // Deprecated keys (will be removed in next major version)
+  /** @deprecated Use __dd_allocation_key instead */
+  allocationKey: string
+  /** @deprecated Not used externally, will be removed */
+  variationType: FlagValueType
+  /** @deprecated Use __dd_do_log instead */
+  doLog: boolean
 }
 
 /**
--- a/packages/core/src/configuration/configuration.ts
+++ b/packages/core/src/configuration/configuration.ts
@@ -56,7 +56,7 @@ export type PrecomputedFlagMetadata = {
   allocationKey: string
   variationType: FlagValueType
   doLog: boolean
-  serialId?: number
+  __dd_split_serial_id?: number
 }
 
 /**
--- a/packages/node-server/package.json
+++ b/packages/node-server/package.json
@@ -36,7 +36,7 @@
     "typecheck": "tsc --noEmit"
   },
   "dependencies": {
-    "@datadog/flagging-core": "0.3.3"
+    "@datadog/flagging-core": "workspace:^"
   },
   "peerDependencies": {
     "@openfeature/server-sdk": ">=1.15.1"
--- a/packages/node-server/src/configuration/evaluateForSubject.ts
+++ b/packages/node-server/src/configuration/evaluateForSubject.ts
@@ -91,7 +91,7 @@ export function evaluateForSubject<T extends FlagValueType>(
             allocationKey: allocation.key,
             variationType: variantTypeToFlagValueType(flag.variationType),
             doLog: !!allocation.doLog,
-            serialId: selectedSplit.serialId,
+            __dd_split_serial_id: selectedSplit.serialId,
           } as PrecomputedFlagMetadata,
         }
       }
--- a/packages/node-server/src/configuration/evaluateForSubject.ts
+++ b/packages/node-server/src/configuration/evaluateForSubject.ts
@@ -88,10 +88,14 @@ export function evaluateForSubject<T extends FlagValueType>(
           reason: StandardResolutionReasons.TARGETING_MATCH,
           variant: variant.key,
           flagMetadata: {
+            // Standardized keys
+            __dd_allocation_key: allocation.key,
+            __dd_do_log: !!allocation.doLog,
+            __dd_split_serial_id: selectedSplit.serialId,
+            // Deprecated keys (for backwards compatibility)
             allocationKey: allocation.key,
             variationType: variantTypeToFlagValueType(flag.variationType),
             doLog: !!allocation.doLog,
-            __dd_split_serial_id: selectedSplit.serialId,
           } as PrecomputedFlagMetadata,
         }
       }
--- a/packages/node-server/src/configuration/evaluateForSubject.ts
+++ b/packages/node-server/src/configuration/evaluateForSubject.ts
@@ -88,11 +88,11 @@ export function evaluateForSubject<T extends FlagValueType>(
           reason: StandardResolutionReasons.TARGETING_MATCH,
           variant: variant.key,
           flagMetadata: {
-            // Standardized keys
+            // Keys for dd-trace-js
             __dd_allocation_key: allocation.key,
             __dd_do_log: !!allocation.doLog,
             __dd_split_serial_id: selectedSplit.serialId,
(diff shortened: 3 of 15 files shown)


packages/node-server/test/data/flags-v1.json
```diff
@@ -59,7 +59,8 @@
               "splits": [
                 {
                   "variationKey": "pi",
-                  "shards": []
+                  "shards": [],
+                  "serialId": 100
```
sameerank: Just added `serialId` to one split because the SDK needs to be ok with either case: if it is present or if it is missing.

packages/node-server/package.json
```diff
@@ -36,7 +36,7 @@
     "typecheck": "tsc --noEmit"
   },
   "dependencies": {
-    "@datadog/flagging-core": "0.3.3"
```
sameerank, line 39: This is referencing a published version, but the tests need the updated type for `flagMetadata` which isn't published yet

packages/core/src/configuration/configuration.ts
```diff
@@ -56,6 +56,7 @@ export type PrecomputedFlagMetadata = {
   allocationKey: string
   variationType: FlagValueType
   doLog: boolean
+  serialId?: number
```
sameerank: I see that `PrecomputedFlagMetadata` is marked as internal, so I assume it's fine to add this field

~`https://github.com/DataDog/openfeature-js-client/blob/a3015e85b1143dfce711cbacecb11af2e283c233/packages/core/src/configuration/configuration.ts#L6-L9`~ (whoops, I showed the wrong `internal` marker)

https://github.com/DataDog/openfeature-js-client/blob/a3015e85b1143dfce711cbacecb11af2e283c233/packages/core/src/configuration/configuration.ts#L54-L55

packages/node-server/src/configuration/evaluateForSubject.ts
```diff
@@ -91,6 +91,7 @@ export function evaluateForSubject<T extends FlagValueType>(
             allocationKey: allocation.key,
             variationType: variantTypeToFlagValueType(flag.variationType),
             doLog: !!allocation.doLog,
+            serialId: selectedSplit.serialId,
```
leoromanovsky: this is potentially undefined; do you want to conditionally add it?

packages/core/src/configuration/configuration.ts
leoromanovsky: Is this just for the client sdk? does it need serial id?

yarn.lock
```diff
@@ -740,7 +731,7 @@ __metadata:
   version: 0.0.0-use.local
   resolution: "@datadog/openfeature-node-server@workspace:packages/node-server"
   dependencies:
-    "@datadog/flagging-core": "npm:0.3.3"
+    "@datadog/flagging-core": "workspace:^"
```
leoromanovsky, line 734: what does this do?

packages/node-server/src/configuration/evaluateForSubject.ts
leoromanovsky: I see in the PR description:

```
The serialId field is optional (serialId?: number). This ensures:

UFC without serialId: If the Universal Flag Configuration does not contain serialId on splits, flagMetadata.serialId will be undefined. Consuming code (tracers) must handle this gracefully.
No breaking changes: Existing integrations continue to work without modification.
The dd-trace-js implementation handles missing serialId gracefully:
```

Let me change my question - is it obvious from the types of `flagMetadata` that the upstream consumer should handle undefined values?

packages/core/src/configuration/configuration.ts
sameerank: This is for the node-server SDK (not the browser client). The `serialId` is needed for APM span enrichment: dd-trace-js reads it from `flagMetadata` and encodes it into span tags (`ffe_flags_enc`)

packages/node-server/src/configuration/evaluateForSubject.ts
sameerank: Yes it's indicated with the `?` in `PrecomputedFlagMetadata` to signal that this field is optional.

https://github.com/DataDog/openfeature-js-client/blob/a3015e85b1143dfce711cbacecb11af2e283c233/packages/core/src/configuration/configuration.ts#L59

The consumer (`SpanEnrichmentHook` in https://github.com/DataDog/dd-trace-js/pull/8343) checks if `(serialId != null)` before using it.

This is for handling missing `serialIds` gracefully (if the UFC is old or if there's a case in the server where a serialId couldn't be provided), so we can still perform assignments even if we don't have everything we need to properly tag spans with feature flag information

yarn.lock
sameerank: This changes the dependency from a fixed npm version (`0.3.3`) to the local workspace package. This is needed so the tests can use the updated type definitions (with `serialId`) that aren't published to npm yet. At publish time, yarn will resolve this to the actual version.

If this change is cause for concern, I could do something like this

```ts
import type { PrecomputedFlagMetadata } from '@datadog/flagging-core'

// Extended type to include serialId for test cases 
// serialId is added in the local workspace version of flagging-core
type TestFlagMetadata = PrecomputedFlagMetadata & {
  serialId?: number
}

export interface TestCase {
  // ...
  result: {
    value: FlagValue
    variant?: string
    flagMetadata?: TestFlagMetadata  // Use extended type
  }
}
```

It's a bit hacky, but avoids touching the dependencies at least

packages/node-server/src/configuration/evaluateForSubject.ts
sameerank: Re:

> do you want to conditionally add it?

I could conditionally spread it: `...(selectedSplit.serialId !== undefined && { serialId: selectedSplit.serialId })` to avoid the key entirely when `undefined` if that's preferred, but that doesn't change the optionality aspect of `serialId` in the type

leoromanovsky: Yea makes sense; upstream right now is dd-trace-js which doesn't have types anyways.

leoromanovsky (approved):

typotter (approved):

packages/core/src/configuration/configuration.ts
petzel: Since this object is really on the `flag` level, consider naming this `allocationSerialId` (similar to how we use `allocationKey`) to make it clear where the serial ID comes from. This would be especially helpful if we were to ever introduce serial IDs for models other than allocations.

sameerank: Oh that's great feedback. Since the serial id lives within each split object in the list of `splits` in the UFC, maybe I could go with `splitSerialId`. I think qualifying with `split` is enough to leave room for ids that map to fewer or more models in the future with other names, e.g. if we add a `variantSerialId` that only maps to variants, and it relates more to the data structure that the SDKs work with (the UFC) vs an internal data model.

The trouble with `allocationSerialId` is 
1. the allocation objects sit one level above the splits where `serialId` is exposed: `"allocations": [{..., "splits": [{..., "serialId": 123}]}]`
2. and we haven't 100% decided where the source of this id will live in our data model

packages/node-server/src/configuration/evaluateForSubject.ts
dd-oleksii: I believe the server SDK is returning this metadata directoly to the use, so consider prefixing the field with `__dd_`

dd-oleksii (approved):

packages/node-server/src/configuration/evaluateForSubject.ts
sameerank: Hmm I wonder if it's odd to do snake case keys. I guess the goal of standardizing to `__dd_allocation_key ` in https://github.com/DataDog/libdatadog/pull/1940 was to make flag metadata easily testable across SDKs, in which case snake vs camel conventions would get in the way

sameerank: I assume we want to be consistent with the prefix for all the internal metadata keys. Is this the right direction? https://github.com/DataDog/openfeature-js-client/pull/273

For the pre-existing non-prefixed keys, the metadata would be updated to include both (w/ and w/o the `__dd_` prefix) and then dd-trace-js would be updated to check for both, and then when we are ready for a major version bump, we can drop the un-prefixed keys

yarn.lock
BridgeAR: This is alright for some dependency managers, while it does not work in all of them and broke npm.

It would have to be a version range 😅 

We have an open range for feature flagging to pull in latest versions automatically, and it broke upstream that way. For now, I will pin the version upstream, as it is difficult to otherwise have the safety for new releases. We could try to run the upstream test suite before a release somehow that integrates the version, but this requires some infrastructure setup (you could have an upstream job that pulls in a specific hash, runs the suite and reports back here). This is the only safe option that immediately comes to my mind for now.

PR #273, opened 2026-05-11
Add standardized __dd_ prefixed keys to flagMetadata

sameerank:
## Motivation

Standardize flagMetadata key names to follow the libdatadog convention (`__dd_` prefix + snake_case). This enables system-tests to validate flag metadata across SDKs.

Related:
* [libdatadog#1940](https://github.com/DataDog/libdatadog/pull/1940)
* https://github.com/DataDog/openfeature-js-client/pull/269/changes/BASE..60b0554f57b44ed75ec4d1fc60394c392b1d0b90#r3218694823

## Changes

- Add `__dd_allocation_key` and `__dd_do_log` as standardized keys in `PrecomputedFlagMetadata`
- Mark `allocationKey`, `doLog`, and `variationType` as `@deprecated`
- Writers (evaluateForSubject.ts, browser/evaluation.ts) set both old and new keys
- Readers (exposureEvent.ts, flagEvaluationAggregator.ts) prefer new keys with fallback to deprecated

## Test instructions

```bash
npm test
```

All 383 tests pass (26 core + 242 node-server + 115 browser).

## Checklist

- [ ] Updated Documentation
- [x] Tested locally
- [ ] Tested on staging

### DIFF
--- a/packages/core/src/configuration/exposureEvent.ts
+++ b/packages/core/src/configuration/exposureEvent.ts
@@ -5,13 +5,15 @@ export function createExposureEvent<T extends FlagValue>(
   context: EvaluationContext,
   details: EvaluationDetails<T>
 ): ExposureEvent | undefined {
-  // Only log if doLog flag is true
-  if (!details.flagMetadata?.doLog) {
+  // Only log if doLog flag is true (prefer new key, fallback to deprecated)
+  const doLog = details.flagMetadata?.__dd_do_log ?? details.flagMetadata?.doLog
+  if (!doLog) {
     return
   }
 
   // Skip logging if allocation key or variant is missing (this should never happen)
-  const allocationKey = details.flagMetadata?.allocationKey as string
+  // Prefer new key, fallback to deprecated
+  const allocationKey = (details.flagMetadata?.__dd_allocation_key ?? details.flagMetadata?.allocationKey) as string
   const variantKey = details.variant
   if (!allocationKey || !variantKey) {
     return
--- a/packages/core/src/configuration/exposureEvent.ts
+++ b/packages/core/src/configuration/exposureEvent.ts
@@ -5,15 +5,13 @@ export function createExposureEvent<T extends FlagValue>(
   context: EvaluationContext,
   details: EvaluationDetails<T>
 ): ExposureEvent | undefined {
-  // Only log if doLog flag is true (prefer new key, fallback to deprecated)
-  const doLog = details.flagMetadata?.__dd_do_log ?? details.flagMetadata?.doLog
-  if (!doLog) {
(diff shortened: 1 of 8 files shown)


packages/core/src/configuration/exposureEvent.ts
```diff
@@ -5,13 +5,15 @@ export function createExposureEvent<T extends FlagValue>(
...
+  // Only log if doLog flag is true (prefer new key, fallback to deprecated)
+  const doLog = details.flagMetadata?.__dd_do_log ?? details.flagMetadata?.doLog
+  if (!doLog) {
     return
   }
 
   // Skip logging if allocation key or variant is missing (this should never happen)
-  const allocationKey = details.flagMetadata?.allocationKey as string
+  // Prefer new key, fallback to deprecated
+  const allocationKey = (details.flagMetadata?.__dd_allocation_key ?? details.flagMetadata?.allocationKey) as string
```
typotter: Consider extracting a helper method since this fallback is used a few times

```suggestion
  const allocationKey = getAllocationKey(details)
```

typotter (approved):
nothing blocking. LGTM

packages/core/src/configuration/exposureEvent.ts
```diff
@@ -5,13 +5,15 @@ export function createExposureEvent<T extends FlagValue>(
   context: EvaluationContext,
   details: EvaluationDetails<T>
 ): ExposureEvent | undefined {
-  // Only log if doLog flag is true
-  if (!details.flagMetadata?.doLog) {
+  // Only log if doLog flag is true (prefer new key, fallback to deprecated)
+  const doLog = details.flagMetadata?.__dd_do_log ?? details.flagMetadata?.doLog
```
dd-oleksii: minor: we know that both are set, so we should drop deprecated usage

dd-oleksii (approved):
not blocking but we don't need fallbacks and can already remove deprecated usage

packages/core/src/configuration/exposureEvent.ts
sameerank: The deprecated keys still need to be written for dd-trace-js, which is an external, independently-released consumer, where we'll need fallbacks for cases when an old openfeature-js-client is being used

But yes, within openfeature-js-client it's not needed. Cleaned up in https://github.com/DataDog/openfeature-js-client/pull/273/commits/a6a4d6a0156ee4654ec6c64949e8ef4174b618ea

packages/core/src/configuration/exposureEvent.ts
sameerank: Removed the fallback logic, so a simple lookup benefits less from a helper

PR #275, opened 2026-05-12
Use canonical FFE fixtures

leoromanovsky:
## Motivation

1. Use `DataDog/ffe-system-test-data` as the canonical FFE/OpenFeature fixture source and keep the shared JavaScript evaluator aligned with the other SDKs. 
2. The original branch had fallen behind `main`, and canonical evaluation exposed a reason mismatch where JavaScript returned `TARGETING_MATCH` for a static integer allocation while Java returned `STATIC`.

## Changes

- Rebase the canonical-fixture integration onto current `main` (`1282b200e0e98ba54201ae90cdce18abf0938f1c`).
- Pin `packages/core/test/ffe-system-test-data` to corrected regex-only rollback head `44092ce8e5448795591fc37ef4441f077b9d34e7` from DataDog/ffe-system-test-data#18.
- Run `ufc-config.json` and the sorted canonical `evaluation-cases/*.json` corpus directly against `@datadog/flagging-core`; keep provider and exposure tests in `packages/node-server`.
- Classify ordinary unsharded allocations as `STATIC`, shard-only allocations as `SPLIT`, rule matches as `TARGETING_MATCH`, and date-gated allocations as `DEFAULT`.
- Ignore malformed flags atomically when allocations, splits, shards, operators, or regexes are invalid, returning the caller default with `DEFAULT`.
- Retain and pass the canonical invalid shard-bound isolation cases, including the negative shard-range fixture.
- Use the canonical fixture corpus as the reason-regression coverage instead of duplicating it in core unit tests.
- Support the retained canonical regex corpus's leading inline flags and POSIX `[:alnum:]` class in JavaScript.
- Initialize the fixture submodule in CI and exclude the external fixture tree from repository formatting and linting.
- Add weekly Dependabot updates for the fixture submodule.


packages/core/test/evaluation/evaluateForSubject.spec.ts
```diff
@@ -92,7 +92,7 @@ describe('evaluateForSubject', () => {
       const result = evaluateForSubject(flag, 'boolean', 'user-123', context, false, logger)
 
       expect(result.value).toBe(true)
-      expect(result.reason).toBe('TARGETING_MATCH')
+      expect(result.reason).toBe('SPLIT')
```
sameerank, line 95: The submodule is providing value immediately on setup 🙌

sameerank (approved):

packages/core/test/evaluation/evaluateForSubject.spec.ts
leoromanovsky: Thank you!

PR #282, opened 2026-05-14
ci: add npm compatibility smoke test

sameerank:
## Motivation

Prevent yarn-specific syntax (like `workspace:^`) from leaking into published npm packages. This would have caught issues where npm users couldn't install packages due to unsupported protocols.

## Changes

Adds a smoke test workflow (`.github/workflows/smoke-test.yaml`) that:
1. Packs each package using `npm pack`
2. Validates no yarn-specific protocols (`workspace:`, `patch:`, `portal:`, `link:`) leak into tarballs
3. Installs packages using npm (not yarn) to catch incompatibilities
4. Verifies packages load correctly and exports are available

**Note:** This PR intentionally includes a commit that introduces `workspace:^` to demonstrate that the smoke test catches the issue. The CI will fail on the second commit, proving the test works.

## Test instructions

1. Watch [CI fail](https://github.com/DataDog/openfeature-js-client/actions/runs/25871643209/job/76028159650?pr=282) on the `workspace:^` commit (second commit)
```
datadog-openfeature-node-server-1.2.1.tgz
    "@datadog/flagging-core": "workspace:^"
❌ Found yarn-specific protocol in packages/node-server
```
2. After observing failure, a fix commit will be pushed
3. Watch CI pass after the fix

## Checklist

- [ ] Updated Documentation
- [ ] Tested locally
- [ ] Tested on staging

### DIFF
--- a/.github/workflows/smoke-test.yaml
+++ b/.github/workflows/smoke-test.yaml
@@ -1,0 +1,94 @@
+name: Package Smoke Test
+
+on:
+  pull_request:
+    paths:
+      - 'packages/*/package.json'
+      - 'yarn.lock'
+      - 'lerna.json'
+
+jobs:
+  npm-compatibility:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+
+      - uses: actions/setup-node@53b83947a5a98c8d113130e565377fae1a50d02f # v6.3.0
+        with:
+          node-version: '24.x'
+          cache: 'yarn'
+
+      - name: Install dependencies
+        run: yarn install --immutable && yarn allow-scripts
+
+      - name: Build packages
+        run: yarn build
+
+      - name: Pack and validate packages
+        run: |
+          echo "Packing packages and checking for yarn-specific protocols..."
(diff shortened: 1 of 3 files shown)


.github/workflows/smoke-test.yaml
```diff
@@ -0,0 +1,97 @@
...
+            # Check for any yarn-specific protocols in the packed package.json
+            if grep -E '"(workspace|patch|portal|link):' /tmp/package/package.json; then
+              echo "❌ Found yarn-specific protocol in $pkg_dir"
+              cat /tmp/package/package.json
+              exit 1
+            fi
+
+            echo "✅ No yarn-specific protocols in $pkg_dir"
+            rm -rf /tmp/package
+            cd - > /dev/null
```
dd-oleksii, line 56: minor: a test install below will catch these issues, so I don't think we need a separate validation

dd-oleksii (approved):

.github/workflows/smoke-test.yaml
sameerank: I think there are benefits to both
1. The grep check is quick, fails early, and displays a clear error message ("Found yarn-specific protocol")
2. The `npm install` acts as a catch-all for other errors and returns a more generic `EUNSUPPORTEDPROTOCOL`

PR #283, opened 2026-05-14
chore: pin exact versions for internal dependencies

sameerank:
## Motivation

Ensure consumers always get the tested combination of package versions rather than potentially mismatched versions due to semver ranges.

With caret (`^`) versions, npm could resolve `@datadog/flagging-core@^1.2.1` to a newer version (e.g., 1.3.0) while `@datadog/openfeature-browser` was built against 1.2.1, potentially causing subtle incompatibilities.

## Changes

- Add `command.version.exact: true` to `lerna.json` - future `yarn release` will use exact versions
- Remove caret from existing `@datadog/flagging-core` dependencies in browser and node-server packages
- Add `scripts/internal-deps-validate.sh` - validates internal deps use exact versions
- Add CI step to run the validation on every PR

## Test instructions

1. Verify CI passes
2. Test that `yarn release` (dry-run) produces exact versions:
   ```bash
   yarn lerna version --no-push --no-git-tag-version minor
   # Check that package.json files show "1.3.0" not "^1.3.0"
   git checkout -- .
   ```
3. Verify the linter catches carets:
   ```bash
   # Temporarily add caret to test
   sed -i '' 's/"@datadog\/flagging-core": "1.2.1"/"@datadog\/flagging-core": "^1.2.1"/' packages/browser/package.json
   ./scripts/internal-deps-validate.sh  # Should fail
   git checkout -- packages/browser/package.json
   ```

## Checklist

- [ ] Updated Documentation
- [x] Tested locally
- [ ] Tested on staging

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -1,4 +1,9 @@
 {
   "npmClient": "yarn",
-  "version": "1.2.1"
+  "version": "1.2.1",
+  "command": {
+    "version": {
+      "exact": true
+    }
+  }
 }
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -45,9 +45,6 @@ jobs:
       - name: Validate package versions
         run: bash scripts/versions-validate.sh
 
-      - name: Validate internal dependencies use exact versions
-        run: bash scripts/internal-deps-validate.sh
-
       - name: Build packages
         run: yarn build
 
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -45,6 +45,9 @@ jobs:
       - name: Validate package versions
         run: bash scripts/versions-validate.sh
 
+      - name: Validate internal dependencies use exact versions
+        run: bash scripts/internal-deps-validate.sh
+
       - name: Build packages
         run: yarn build
 
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -29,6 +29,9 @@ jobs:
 
           echo "✅ Release validation passed"
 
+      - name: Validate internal dependencies use exact versions
+        run: bash scripts/internal-deps-validate.sh
+
   build-and-publish:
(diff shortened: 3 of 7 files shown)


lerna.json
```diff
@@ -1,4 +1,9 @@
 {
   "npmClient": "yarn",
-  "version": "1.2.1"
+  "version": "1.2.1",
+  "command": {
+    "version": {
+      "exact": true
```
dd-oleksii, line 6: 👍 oh, I thinks that's why it converted references to carets [in the last bump](https://github.com/DataDog/openfeature-js-client/commit/2e3452ee73fdc3997288b23549e9c7261419d9f7) — I forgot `--exact` 😢

fwiw, I think this is the only change that we actually need. lerna will pin versions on the next release, so it doesn't matter what we  currently have in package.json and ci validation is checking irrelevant code.

If you do want to keep validation, I'd put it in pre-release workflow instead, so it checks the source we attempt to release

dd-oleksii (approved):

lerna.json
dd-oleksii: also, because we already published a version with caret dependencies, the safest move is to do a major bump, so that released version doesn't start picking up unmatched dependency versions

sameerank: > lerna will pin versions on the next release, so it doesn't matter what we currently have in package.json

Removed the package.json changes in https://github.com/DataDog/openfeature-js-client/pull/283/commits/5357491b3703dc3746364db86c837cc0f0af2ffe we should see the version get pinned exactly when we run through the release process next time

> ci validation is checking irrelevant code

Moved to the pre-release workflow in https://github.com/DataDog/openfeature-js-client/pull/283/commits/bc8b6cd336fbbbc8ce7f6d4a288735ce1533de11 

> we already published a version with caret dependencies, the safest move is to do a major bump

A major bump makes sense. Safer to establish a clean baseline with exact pinning in 2.0.0

PR #287, opened 2026-05-20
v2.0.0

sameerank:
### Breaking Changes

The `@datadog/openfeature-node-server` and `@datadog/openfeature-browser` packages now declare **exact versions** for their internal dependency on `@datadog/flagging-core`:

| Package | Before (v1.2.1) | After (v2.0.0) |
|---------|-----------------|----------------|
| `@datadog/openfeature-node-server` | `"@datadog/flagging-core": "^1.2.1"` | `"@datadog/flagging-core": "2.0.0"` |
| `@datadog/openfeature-browser` | `"@datadog/flagging-core": "^1.2.1"` | `"@datadog/flagging-core": "2.0.0"` |

### Why a Major Bump?

Users of `dd-trace-js` v5.89.0 through v5.102.0 have `"@datadog/openfeature-node-server": "^1.1.x"` in their dependency tree, which currently resolves to v1.2.1 with loose internal `^` versioning.

A major version ensures these users must explicitly opt-in to the new versioning scheme and won't accidentally receive future flagging-core updates through the loose caret constraint.

### dd-trace-js Dependency History

| dd-trace version | openfeature-node-server dep | Notes |
|------------------|---------------------------|-------|
| v5.72.0 - v5.78.0 | `"0.1.0-preview.x"` | Exact (pinned) |
| v5.79.0 - v5.88.0 | `"^0.2.0"` to `"^0.3.3"` | Caret, but won't resolve to 1.x |
| **v5.89.0 - v5.102.0** | `"^1.1.0"` to `"^1.1.2"` | **Caret - resolves to 1.2.1** |
| v5.103.0+ | `"1.1.2"` | Exact (pinned) |

### Changes

- chore: pin exact versions for internal dependencies ([#283](https://github.com/DataDog/openfeature-js-client/pull/283))
- ci: add npm compatibility smoke test ([#282](https://github.com/DataDog/openfeature-js-client/pull/282))

### Code Diff

https://github.com/DataDog/openfeature-js-client/compare/v1.2.1...v2.0.0

### Checklist

- [x] Version bumped via `yarn release`
- [x] CHANGELOG.md updated
- [ ] Create GitHub Release after merge (tag: `v2.0.0`)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,27 @@
 
 ---
 
+## v2.0.0
+
+**Breaking Changes:**
+
+- chore: pin exact versions for internal dependencies ([#283](https://github.com/DataDog/openfeature-js-client/pull/283))
+
+  The `@datadog/openfeature-node-server` and `@datadog/openfeature-browser` packages now declare **exact versions** for their internal dependency on `@datadog/flagging-core`:
+
+  | Package                            | Before (v1.2.1)                      | After (v2.0.0)                      |
+  | ---------------------------------- | ------------------------------------ | ----------------------------------- |
+  | `@datadog/openfeature-node-server` | `"@datadog/flagging-core": "^1.2.1"` | `"@datadog/flagging-core": "2.0.0"` |
+  | `@datadog/openfeature-browser`     | `"@datadog/flagging-core": "^1.2.1"` | `"@datadog/flagging-core": "2.0.0"` |
+
+  **Why a major bump?**
+
+  Users of `dd-trace-js` v5.89.0 through v5.102.0 have `"@datadog/openfeature-node-server": "^1.1.x"` in their dependency tree, which resolves to v1.2.1 with loose internal versioning. A major version ensures these users must explicitly opt-in to the new versioning scheme.
+
+**Internal Changes:**
+
+- ci: add npm compatibility smoke test ([#282](https://github.com/DataDog/openfeature-js-client/pull/282))
+
 ## v1.2.0
 
 **Internal Changes:**
--- a/lerna.json
+++ b/lerna.json
@@ -1,6 +1,6 @@
 {
   "npmClient": "yarn",
-  "version": "1.2.1",
+  "version": "2.0.0",
   "command": {
     "version": {
       "exact": true
--- a/packages/browser/package.json
+++ b/packages/browser/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@datadog/openfeature-browser",
-  "version": "1.2.1",
+  "version": "2.0.0",
   "description": "Browser-specific bindings for OpenFeature (wraps @datadog/openfeature-core)",
   "license": "Apache-2.0",
   "homepage": "https://github.com/DataDog/openfeature-js-client#readme",
@@ -45,7 +45,7 @@
   },
   "dependencies": {
     "@datadog/browser-core": "^6.33.0",
-    "@datadog/flagging-core": "^1.2.1"
+    "@datadog/flagging-core": "2.0.0"
   },
   "peerDependencies": {
     "@openfeature/core": "^1.9.2",
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@datadog/flagging-core",
-  "version": "1.2.1",
+  "version": "2.0.0",
   "description": "Runtime-agnostic flag-evaluation logic for Datadog Flagging",
   "license": "Apache-2.0",
   "repository": {
--- a/packages/node-server/package.json
+++ b/packages/node-server/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@datadog/openfeature-node-server",
-  "version": "1.2.1",
+  "version": "2.0.0",
   "description": "Node.js server bindings for OpenFeature (wraps @datadog/flagging-core)",
   "license": "Apache-2.0",
   "homepage": "https://github.com/DataDog/openfeature-js-client#readme",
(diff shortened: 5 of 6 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,27 @@
...
+  The `@datadog/openfeature-node-server` and `@datadog/openfeature-browser` packages now declare **exact versions** for their internal dependency on `@datadog/flagging-core`:
+
+  | Package                            | Before (v1.2.1)                      | After (v2.0.0)                      |
+  | ---------------------------------- | ------------------------------------ | ----------------------------------- |
+  | `@datadog/openfeature-node-server` | `"@datadog/flagging-core": "^1.2.1"` | `"@datadog/flagging-core": "2.0.0"` |
+  | `@datadog/openfeature-browser`     | `"@datadog/flagging-core": "^1.2.1"` | `"@datadog/flagging-core": "2.0.0"` |
+
+  **Why a major bump?**
+
+  Users of `dd-trace-js` v5.89.0 through v5.102.0 have `"@datadog/openfeature-node-server": "^1.1.x"` in their dependency tree, which resolves to v1.2.1 with loose internal versioning. A major version ensures these users must explicitly opt-in to the new versioning scheme.
```
greghuels, line 24: I don't think this meets SEMVER's criteria for a major change. The pinned versions are within the previous version ranges, so we don't need to signal breaking API compatibility. I'd use 1.3.0 for this.

sameerank: The caret in the dependency unfortunately means that a minor bump isn't safe for internally consistent versions

With v1.3.0 release (`flagging-core` 1.3.0 also published):
- User stays on `openfeature-node-server` 1.2.1 (lockfile)
- v1.2.1 has `"@datadog/flagging-core": "^1.2.1"` (loose)
- `^1.2.1` resolves to `flagging-core` 1.3.0
- Result: `openfeature-node-server` 1.2.1 + `flagging-core` 1.3.0 (mismatch)

With v2.0.0 release (`flagging-core` 2.0.0 also published):
- User stays on `openfeature-node-server` 1.2.1 (lockfile)
- v1.2.1 has `"@datadog/flagging-core": "^1.2.1"` (loose)
- ^1.2.1 does NOT match 2.0.0 (major bump)
- Result: `openfeature-node-server` 1.2.1 + `flagging-core` 1.2.1 (consistent)

sameerank: Perhaps a question for @typotter who added lerna in https://github.com/DataDog/openfeature-js-client/pull/9:

We are currently internally fixing versioning (all packages share the same version). Are you open to us switching to independent versioning? https://lerna.js.org/docs/features/version-and-publish#versioning-strategies

- `@datadog/flagging-core` - stay at 1.2.1 (no new release)
- `@datadog/openfeature-node-server` - 1.3.0 with `"@datadog/flagging-core": "1.2.1"` (pinned)
- `@datadog/openfeature-browser` - 1.3.0 with `"@datadog/flagging-core": "1.2.1"` (pinned)

This seems cleaner, with the tradeoff being that we have to do independent versioning going forward, but the docs indicate that we could just leave this to lerna by updating the `lerna.json`

```
{
  "version": "independent",
  "command": { "version": { "exact": true } }
}
```

greghuels: I think a v2.0.0 release is better than independent versioning. There are a lot of advantages to ensuring versions stay consistent when considering release notes and understanding which features are available in which versions.

greghuels (approved):
You might want to see if there's anything else breaking that might be a good opportunity to add to this release. But either way, I'm good with using v2.0.0.

CHANGELOG.md
dd-oleksii: Client and server libraries have a disjoint set of features, so having a shared version does not help at all. Client and server libraries also have different status: client is a public library, while server one is an internal'ish dependency of dd-trace. Having a shared version requires us bumping client version when nothing changed, including causing users go through a major bump for no reason.

greghuels: > Client and server libraries have a disjoint set of features

If we're going to treat them as disjoint, we might consider separate repositories with separate release notes. The reason I kept node in the same repo is because if we ever add more advanced stats features like contextual bandits, there would be a lot of code sharing happening between the libraries.

dd-oleksii: > if we ever add more advanced stats features like contextual bandits, there would be a lot of code sharing happening between the libraries

Bandits would get precomputed on edge. I don't think we'll get much sharing until we add dynamic evaluation to client sdk

typotter: > Are you open to us switching to independent versioning?

Yes, absolutely if it makes things easier. The fixed versioning is an artifact of moving fast early on and copying a lot from how the rum SDK package hierarchy is set up.

greghuels: It looks like [emotion's tagging system](https://github.com/emotion-js/emotion/tags) would keep things organized if we choose to go down that route.

sameerank: The consensus sounds like we’re open to trying independent versioning.

At the moment, I think it would be preferable to not force a major bump on customers for a bug fix. Also this approach to pin the `flagging-core` dependency would only be temporary. Any future changes to `flagging-core` will require bumping it to `2.0.0`, since users on `openfeature-node-server@1.2.1` would still pull new 1.x versions via the `^1.2.1` constraint. However, we can at least save the major bump for a more appropriate situation.

I’m going to 
1. close this PR
2. open a new PR for an independent versioning (including emotion-style tags) release workflow https://github.com/DataDog/openfeature-js-client/pull/288
3. and then we can do a 1.3.0 release with independent versioning

PR #292, opened 2026-05-22
Release openfeature-browser and openfeature-node-server v1.3.0

sameerank:
## Summary

Pin internal `@datadog/flagging-core` dependency to exact version `1.2.1` and bump `@datadog/openfeature-browser` and `@datadog/openfeature-node-server` to v1.3.0.

## Motivation

This is the first stable release with the serialId feature:
- **1.2.0**: Broken due to yarn syntax issues
- **1.2.1**: Introduced loose dependencies (`^1.2.1`) which prevents safely releasing updates to flagging-core
- **1.3.0**: First stable release with exact dependency pinning

## Changes

- `packages/browser/package.json`: Pin `@datadog/flagging-core` to `1.2.1`, bump version to `1.3.0`
- `packages/node-server/package.json`: Pin `@datadog/flagging-core` to `1.2.1`, bump version to `1.3.0`
- `yarn.lock`: Updated to reflect exact versions
- `CHANGELOG.md`: Added v1.3.0 entry

## Next Steps

After merge, create GitHub releases for:
- `@datadog/openfeature-browser@1.3.0`
- `@datadog/openfeature-node-server@1.3.0`

### DIFF
--- a/packages/browser/package.json
+++ b/packages/browser/package.json
@@ -45,7 +45,7 @@
   },
   "dependencies": {
     "@datadog/browser-core": "^6.33.0",
-    "@datadog/flagging-core": "^1.2.1"
+    "@datadog/flagging-core": "1.2.1"
   },
   "peerDependencies": {
     "@openfeature/core": "^1.9.2",
--- a/packages/browser/package.json
+++ b/packages/browser/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@datadog/openfeature-browser",
-  "version": "1.2.1",
+  "version": "1.3.0",
   "description": "Browser-specific bindings for OpenFeature (wraps @datadog/openfeature-core)",
   "license": "Apache-2.0",
   "homepage": "https://github.com/DataDog/openfeature-js-client#readme",
--- a/packages/node-server/package.json
+++ b/packages/node-server/package.json
@@ -36,7 +36,7 @@
     "typecheck": "tsc --noEmit"
   },
   "dependencies": {
-    "@datadog/flagging-core": "^1.2.1"
+    "@datadog/flagging-core": "1.2.1"
   },
   "peerDependencies": {
     "@openfeature/server-sdk": ">=1.15.1"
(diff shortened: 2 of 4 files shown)


packages/browser/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@datadog/openfeature-browser",
-  "version": "1.2.1",
+  "version": "1.3.0",
```
sameerank, line 3: Now with independent versioning, it's not strictly required to bump browser too, but it seemed like a good idea to publish a version without the `"@datadog/flagging-core": "^1.2.1"` dependency

packages/node-server/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@datadog/openfeature-node-server",
-  "version": "1.2.1",
+  "version": "1.3.0",
```
dd-oleksii, line 3: we need 2.0.0 for server, no?

packages/browser/package.json
dd-oleksii: 1.2.2 is better for this than 1.3.0

packages/node-server/package.json
sameerank: Hmm, I thought the concern was with `"@datadog/flagging-core": "^1.2.1"`: that the next release of `flagging-core` will need to be a major version to ensure no impact to anyone on `node-server@1.2.1` with the loose dependency to `flagging-core` in their lock file. And independent versioning frees us from having to do that update since we can pin `flagging-core` without also needing to also increment its version to satisfy the old approach of uniform fixed versioning.

It doesn't hurt to do a major bump to `node-server` since it's internal, although I'm having trouble understanding the reason for it 🤔

packages/browser/package.json
sameerank: With my understanding of how independent versioning works, I think `node-server` could be a patch update too

sameerank:
Based on the feedback above, closing in favor of https://github.com/DataDog/openfeature-js-client/pull/293

packages/node-server/package.json
dd-oleksii: Because old dd-trace versions have unpinned dependency on @datadog/openfeature-node-server

PR #317, opened 2026-06-22
Align Node.js flagevaluation EVP metadata

leoromanovsky:
## Motivation

Node.js server-side flag-evaluation EVP delivery depends on the OpenFeature node-server package providing accurate evaluation metadata to `dd-trace-js`. Customers need Node.js evaluations to land in the same backend `flagevaluation` signal as the other SDKs, and APM needs that signal to reflect evaluation time, runtime-default behavior, and bounded context consistently across runtimes.

## Changes

- Adds `__dd_eval_timestamp_ms` evaluation-entry metadata to node-server resolution details so EVP `first_evaluation` and `last_evaluation` come from evaluation time rather than flush time.
- Uses js-core `timeStampNow()` for typed evaluation and outbound EVP envelope timestamps.
- Treats runtime default as absent variant, not OpenFeature `reason` or `errorCode` alone.
- Keeps targeting key as the dedicated EVP field and prevents duplicate `targetingKey`/`targeting_key` entries in `context.evaluation`.
- Allows error/default metadata to carry only internal tracing metadata when no allocation metadata exists.
- Renames the browser flag-evaluation hook implementation from generic tracking language to `createFlagEvalEVPHook`, matching the cross-SDK metrics-vs-EVP split.
- Adds coverage for timestamp metadata, runtime-default semantics, and provider hook visibility.
- Stabilizes affected timer tests around Jest system time.

## Decisions

- `__dd_eval_timestamp_ms` uses the existing `__dd_*` internal metadata namespace for SDK-to-tracer coordination; it is not a public API promise.
- Runtime-default derivation follows the provider result shape: an absent variant means the SDK returned the caller default. OpenFeature `reason` and `errorCode` are not used as independent runtime-default truth because they are not part of the flagevaluation EVP cardinality contract.
- The TypeScript aggregator keeps the runtime-default rule local and schema-oriented. Type mismatch remains an OpenFeature resolution detail, but this provider's type-mismatch path already reaches the aggregator without a variant when it uses the caller default.
- Browser hook implementation names use the same metrics-vs-EVP split as the tracer SDKs; payload, event, and aggregator types keep `FlagEvaluation*` names because those describe the transport contract rather than a hook role.
- Error/default resolution metadata can be partial because those paths need to preserve evaluation timing even when no allocation, variation type, or exposure `doLog` metadata exists.
- This is a companion to the `dd-trace-js` Node.js EVP writer draft; the writer remains in the tracer repo while this package supplies canonical evaluation metadata.

### Dogfooding App

- Companion Node.js tracer draft: https://github.com/DataDog/dd-trace-js/pull/8902
- `ffe-dogfooding` `app-nodejs` was run with local `dd-trace-js`, `@datadog/flagging-core`, and `@datadog/openfeature-node-server` artifacts and reached `PROVIDER_READY`.
- Evaluated `ffe-dogfooding-string-flag` five times for each public-safe targeting key, keeping the evaluation context stable per key so batching/aggregation is observable:
  - `nodejs-batch-evp-agent-20260623T024835Z-alpha`
  - `nodejs-batch-evp-agent-20260623T024835Z-bravo`
  - `nodejs-batch-evp-agent-20260623T024835Z-charlie`
- App-side result: all 15 evaluations returned `variant_2` with `reason=TARGETING_MATCH`.

### System Tests

- Companion draft PR: https://github.com/DataDog/system-tests/pull/7182

### Staging End-To-End

- Dogfooding ran without the local mock-intake EVP tee/proxy, so the Agent sent EVP traffic through the normal backend path.
- Retriever staging query used `eventplatform.system.track(TRACK => 'flagevaluation')` against `us1.staging.dog` for the exact targeting keys above.
- Backend rows observed:
  - `nodejs-batch-evp-agent-20260623T024835Z-alpha`: `evaluation_count=5`, `flag.key=ffe-dogfooding-string-flag`, `variant.key=variant_2`, `allocation.key=allocation-override-392dd7c149f8`
  - `nodejs-batch-evp-agent-20260623T024835Z-bravo`: `evaluation_count=5`, `flag.key=ffe-dogfooding-string-flag`, `variant.key=variant_2`, `allocation.key=allocation-override-392dd7c149f8`
  - `nodejs-batch-evp-agent-20260623T024835Z-charlie`: `evaluation_count=5`, `flag.key=ffe-dogfooding-string-flag`, `variant.key=variant_2`, `allocation.key=allocation-override-392dd7c149f8`

### DIFF
--- a/packages/core/src/configuration/configuration.ts
+++ b/packages/core/src/configuration/configuration.ts
@@ -49,6 +49,8 @@ export type PrecomputedFlag<T extends FlagValueType = FlagValueType> = {
 
 /** @internal */
 export type PrecomputedFlagMetadata = {
+  // Metadata carried on OpenFeature evaluation details. The type name is kept
+  // for package API compatibility; server fallback/error results also use it.
   // Primary keys (used by browser and core readers)
   allocationKey?: string
   variationType?: FlagValueType
--- a/packages/core/src/configuration/configuration.ts
+++ b/packages/core/src/configuration/configuration.ts
@@ -62,6 +62,7 @@ export type PrecomputedFlagMetadata = {
   __dd_allocation_key?: string
   __dd_do_log?: boolean
   __dd_split_serial_id?: number
+  'dd.eval.timestamp_ms'?: UnixTimestamp
 }
 
 /**
--- a/packages/core/src/configuration/flagEvaluationAggregator.ts
+++ b/packages/core/src/configuration/flagEvaluationAggregator.ts
@@ -118,7 +118,7 @@ export class FlagEvaluationAggregator {
 
 function getEvaluationTimestamp<T extends FlagValue>(details: EvaluationDetails<T>): TimeStamp {
   const metadataTimestamp = details.flagMetadata?.[EVALUATION_TIMESTAMP_METADATA_KEY]
-  return typeof metadataTimestamp === 'number' ? (metadataTimestamp as TimeStamp) : timeStampNow()
+  return Number.isFinite(metadataTimestamp) ? (metadataTimestamp as TimeStamp) : timeStampNow()
 }
 
 function isRuntimeDefaultUsed<T extends FlagValue>(details: EvaluationDetails<T>): boolean {
--- a/packages/core/src/configuration/flagEvaluationAggregator.ts
+++ b/packages/core/src/configuration/flagEvaluationAggregator.ts
@@ -3,6 +3,8 @@ import { getMD5Hash } from '../obfuscation'
 import { createFlagEvaluationEvent } from './flagEvaluationEvent'
 import type { FlagEvaluationEvent } from './flagEvaluationEvent.types'
 
+const EVALUATION_TIMESTAMP_METADATA_KEY = 'dd.eval.timestamp_ms'
+
 interface FlagEvaluationAggregationData {
   flagKey: string
   variantKey?: string
@@ -44,7 +46,7 @@ export class FlagEvaluationAggregator {
 
   addEvaluation<T extends FlagValue>(context: EvaluationContext, details: EvaluationDetails<T>, error?: string): void {
     const keyString = this.createAggregationKeyString(context, details, error)
-    const timestamp = Date.now()
+    const timestamp = getEvaluationTimestamp(details)
 
     const existingData = this.aggregatedData.get(keyString)
     if (existingData) {
@@ -54,7 +56,7 @@ export class FlagEvaluationAggregator {
         existingData.error = error
       }
     } else {
-      const runtimeDefaultUsed = details.reason === 'DEFAULT' || details.reason === 'ERROR'
+      const runtimeDefaultUsed = details.variant == null || details.errorCode === 'TYPE_MISMATCH'
       const allocationKey = details.flagMetadata?.allocationKey as string
       const targetingRuleKey = details.flagMetadata?.targetingRuleKey as string
       const { targetingKey, ...targetingContext } = context
@@ -80,8 +82,9 @@ export class FlagEvaluationAggregator {
       return
     }
 
+    const flushTimestamp = Date.now()
     const events = Array.from(this.aggregatedData.values()).map((data) =>
-      createFlagEvaluationEvent(data, data.firstEvaluation)
+      createFlagEvaluationEvent(data, flushTimestamp)
     )
     this.aggregatedData.clear()
     this.onFlush(events)
@@ -109,4 +112,9 @@ export class FlagEvaluationAggregator {
       })
     )
   }
+}
+
+function getEvaluationTimestamp<T extends FlagValue>(details: EvaluationDetails<T>): number {
+  const metadataTimestamp = details.flagMetadata?.[EVALUATION_TIMESTAMP_METADATA_KEY]
+  return typeof metadataTimestamp === 'number' ? metadataTimestamp : Date.now()
 }
--- a/packages/core/src/configuration/flagEvaluationAggregator.ts
+++ b/packages/core/src/configuration/flagEvaluationAggregator.ts
@@ -122,5 +122,10 @@ function getEvaluationTimestamp<T extends FlagValue>(details: EvaluationDetails<
 }
 
 function isRuntimeDefaultUsed<T extends FlagValue>(details: EvaluationDetails<T>): boolean {
+  // Datadog-assigned evaluations attach a platform variation key to OpenFeature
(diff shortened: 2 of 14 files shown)


packages/core/src/configuration/configuration.ts
```diff
@@ -58,6 +62,7 @@ export type PrecomputedFlagMetadata = {
   __dd_allocation_key?: string
   __dd_do_log?: boolean
   __dd_split_serial_id?: number
+  'dd.eval.timestamp_ms'?: UnixTimestamp
```
leoromanovsky: this should be `__dd` ?

packages/core/src/configuration/flagEvaluationAggregator.ts
```diff
@@ -56,7 +56,7 @@ export class FlagEvaluationAggregator {
         existingData.error = error
       }
     } else {
-      const runtimeDefaultUsed = details.reason === 'DEFAULT' || details.reason === 'ERROR'
+      const runtimeDefaultUsed = details.variant == null || details.errorCode === 'TYPE_MISMATCH'
```
leoromanovsky: what's going on here?

packages/core/src/configuration/flagEvaluationAggregator.ts
```diff
@@ -112,3 +115,12 @@ export class FlagEvaluationAggregator {
...
   }
 }
+
+function getEvaluationTimestamp<T extends FlagValue>(details: EvaluationDetails<T>): TimeStamp {
+  const metadataTimestamp = details.flagMetadata?.[EVALUATION_TIMESTAMP_METADATA_KEY]
+  return typeof metadataTimestamp === 'number' ? (metadataTimestamp as TimeStamp) : timeStampNow()
+}
+
+function isRuntimeDefaultUsed<T extends FlagValue>(details: EvaluationDetails<T>): boolean {
+  return details.variant == null
```
typotter, line 130: Do we have an explicit contract that the platform-created variants can never be null?

typotter (commented):
LGTM for the timestamps, but question the `variant==null` therefore runtime default change.
Is the existing ERROR reason check in-exhaustive or too wide? We consider it an error when the waterfall fails to assign a variant including flag not found, actual errors encountered, and a lack of _some_ allocation that matches.

packages/core/src/configuration/flagEvaluationAggregator.ts
leoromanovsky: Good question. The SDK-visible contract here is that assigned Datadog evaluations attach an OpenFeature `details.variant` from the platform variation key: browser precomputed flags model `variationKey` as a string, and server UFC variants are keyed by string. Default/error fallback paths omit `variant`, so nullish `variant` is the runtime-default signal; `reason`/`errorCode` alone is intentionally not enough.

I pushed 27d292f to make that explicit in the helper comment and extended the test matrix so `TYPE_MISMATCH`/general error with an assigned variant stays `runtime_default_used=false`.

Validation:
- `yarn workspace @datadog/flagging-core test --runTestsByPath test/configuration/flagEvaluationAggregator.spec.ts && yarn workspace @datadog/flagging-core typecheck`
- `yarn workspace @datadog/openfeature-browser test --runTestsByPath test/openfeature/flagEvaluations.spec.ts test/openfeature/provider.spec.ts && yarn workspace @datadog/openfeature-browser typecheck`

dd-oleksii: `details.variant` is variant key (not value), so yes

leoromanovsky: Originally the agent wanted to include other conditions but I think omission of variant is best; it's up to us to make sure this is always nullish in exceptional situations, for example when there is a type mismatch.

packages/core/src/configuration/configuration.ts
```diff
@@ -50,14 +50,15 @@ export type PrecomputedFlag<T extends FlagValueType = FlagValueType> = {
 /** @internal */
 export type PrecomputedFlagMetadata = {
   // Primary keys (used by browser and core readers)
-  allocationKey: string
-  variationType: FlagValueType
-  doLog: boolean
+  allocationKey?: string
+  variationType?: FlagValueType
+  doLog?: boolean
```
sameerank, line 57: A thought: the name `PrecomputedFlagMetadata` implies a successfully precomputed flag. If these are absent, it's typically due to not not succeeding, e.g. if a flag is missing. Probably too late to rename the type, but it would have been nice to call it something like `FlagEvaluationMetadata`. Or another idea might be to have two types, one that is strict for successes and one that is more flexible for fallback/error metadata

packages/node-server/src/configuration/evaluateForSubject.ts
```diff
@@ -117,9 +123,14 @@ export function evaluateForSubject<T extends FlagValueType>(
...
     value: defaultValue,
     reason: StandardResolutionReasons.DEFAULT,
+    flagMetadata: createEvaluationTimestampMetadata(evaluationTimestampMs),
   }
 }
 
+function createEvaluationTimestampMetadata(evaluationTimestampMs: TimeStamp): PrecomputedFlagMetadata {
+  return { __dd_eval_timestamp_ms: evaluationTimestampMs } as PrecomputedFlagMetadata
+}
+
```
sameerank: Probably could be a shared helper

https://github.com/DataDog/openfeature-js-client/blob/e412bcda22788895a823b2db0925976fe90a5ab1/packages/node-server/src/configuration/evaluation.ts#L83-L85

packages/core/src/configuration/flagEvaluationAggregator.ts
```diff
@@ -112,3 +115,17 @@ export class FlagEvaluationAggregator {
     )
   }
 }
+
+function getEvaluationTimestamp<T extends FlagValue>(details: EvaluationDetails<T>): TimeStamp {
+  const metadataTimestamp = details.flagMetadata?.[EVALUATION_TIMESTAMP_METADATA_KEY]
+  return typeof metadataTimestamp === 'number' ? (metadataTimestamp as TimeStamp) : timeStampNow()
```
sameerank:
```suggestion
  return Number.isFinite(metadataTimestamp) ? (metadataTimestamp as TimeStamp) : timeStampNow()
```

Also protects against `NaN`, which is a number despite its name

sameerank (approved):
Just mentioned a few nits, but nothing blocking

PR #318, opened 2026-06-26
feat(browser): add DevToolsProvider and initFeatureFlags for override…

kellyw1806:
## Context

The Datadog Configurator currently applies flag overrides through wrapper functions in web-ui that short-circuit before OpenFeature runs. This means hooks never fire for overridden flags, there's no single observation point to record what flags a page evaluated, and customers using the SDK directly have no override mechanism at all.

This PR moves overrides inside the OpenFeature chain using `MultiProvider`, so the core provider stays untouched and the override mechanism is available to all consumers.

## What changed

- **`DevToolsProvider`** — a new provider that reads from `localStorage['dd.dd_flag.overrides']`. On override hit, returns the value with `flagMetadata.overridden = true` and `reason: STATIC`. On miss, throws `FlagNotFoundError` so `MultiProvider` falls through to `DatadogProvider` normally. Includes runtime type validation — throws `TypeMismatchError` if the declared type and actual value type don't match (e.g. `{ type: 'BOOLEAN', value: 'yes' }` or `{ type: 'INTEGER', value: 3.14 }`)
- **Tests** — full coverage across all 5 flag types (BOOLEAN, STRING, INTEGER, NUMERIC, JSON): override hit, miss, type mismatch, malformed localStorage, multiple coexisting overrides

## Consumer usage

Consumers opt in by composing DevToolsProvider in front of DatadogProvider via MultiProvider. Dev tools off (default) means the provider is absent from the chain entirely — no bundle impact.

```js
// dev tools on
OpenFeature.setProvider(new MultiProvider([
  { provider: new DevToolsProvider(), name: 'devtools' },
  { provider: new DatadogProvider(config), name: 'datadog' },
]))

// dev tools off (default)
OpenFeature.setProvider(new DatadogProvider(config))
```

Overrides live in localStorage['dd.dd_flag.overrides'] as { type, value } pairs. Overridden evaluations come back with flagMetadata.overridden: true.

## Test plan

- [ ] Override hit: set dd.dd_flag.overrides, evaluate flag — resolves to override value, not the Datadog-configured value
- [ ] Override miss: no entry — falls through to DatadogProvider and resolves normally
- [ ] Type mismatch: { type: 'BOOLEAN', value: 'yes' } throws TypeMismatchError
- [ ] INTEGER float: { type: 'INTEGER', value: 3.14 } throws TypeMismatchError
- [ ] enableDevTools: false (default): no MultiProvider, no DevToolsProvider, marker absent
- [ ] enableDevTools: true: dd.dd_flag.devtools = "enabled" present in localStorage after init
- [ ] Run yarn workspace @datadog/openfeature-browser test — all 20 tests pass


packages/browser/src/init.ts
```diff
@@ -0,0 +1,32 @@
...
+export interface InitFeatureFlagsOptions extends FlaggingInitConfiguration {
+  enableDevTools?: boolean
+  userContext?: EvaluationContext
+  orgContext?: EvaluationContext
+}
+
+export async function initFeatureFlags({
+  enableDevTools = false,
+  userContext = {},
+  orgContext = {},
```
leoromanovsky: "org" is a datadog specific context

packages/browser/src/openfeature/devtools-provider.ts
```diff
@@ -0,0 +1,99 @@
...
+  Provider,
+  ProviderMetadata,
+  ResolutionDetails,
+} from '@openfeature/web-sdk'
+import { FlagNotFoundError } from '@openfeature/web-sdk'
+
+const OVERRIDES_KEY = 'dd.dd_flag.overrides'
+const DEVTOOLS_MARKER_KEY = 'dd.dd_flag.devtools'
+
+type DDFlagOverrideType = 'BOOLEAN' | 'STRING' | 'INTEGER' | 'NUMERIC' | 'JSON'
```
leoromanovsky, line 15: possible to use existing types?

packages/browser/src/openfeature/devtools-provider.ts
```diff
@@ -0,0 +1,99 @@
...
+
+interface DDFlagOverride {
+  type: DDFlagOverrideType
+  value: boolean | string | number | JsonValue
+  variants?: Record<string, string>
+}
+
+function readOverrides(): Record<string, DDFlagOverride> {
+  try {
+    return JSON.parse(localStorage.getItem(OVERRIDES_KEY) ?? '{}') as Record<string, DDFlagOverride>
```
leoromanovsky, line 25: we moved away from using local storage in favor of indexeddb; consider whether that API meets your needs.

packages/browser/src/openfeature/devtools-provider.ts
```diff
@@ -0,0 +1,99 @@
...
+/**
+ * A dev-only provider that applies flag overrides from localStorage (key: dd.dd_flag.overrides).
+ * Intended to be composed in front of the real provider via MultiProvider + FirstMatchStrategy:
+ *
+ *   OpenFeature.setProvider(new MultiProvider([
+ *     { provider: new DevToolsProvider(), name: 'devtools' },
+ *     { provider: new DatadogProvider(...), name: 'datadog' },
+ *   ]))
+ *
+ * On override hit: returns the override value with reason STATIC.
```
leoromanovsky, line 66: Good choice; please add tests to validate

packages/browser/src/openfeature/devtools-provider.ts
```diff
@@ -0,0 +1,99 @@
...
+}
+
+/**
+ * A dev-only provider that applies flag overrides from localStorage (key: dd.dd_flag.overrides).
+ * Intended to be composed in front of the real provider via MultiProvider + FirstMatchStrategy:
+ *
+ *   OpenFeature.setProvider(new MultiProvider([
+ *     { provider: new DevToolsProvider(), name: 'devtools' },
+ *     { provider: new DatadogProvider(...), name: 'datadog' },
+ *   ]))
```
leoromanovsky, line 64: Please update the readme in this PR to reflect how customers will use this.

leoromanovsky (changes requested):

packages/browser/src/openfeature/devtools-provider.ts
```diff
@@ -0,0 +1,99 @@
...
+ *
+ * On override hit: returns the override value with reason STATIC.
+ * On miss: throws FlagNotFoundError so MultiProvider falls through to the next provider.
+ *
+ * Flag recording (discovery) is handled separately by createEvaluatedFlagsHook(), which
```
sameerank: Not seeing `createEvaluatedFlagsHook` anywhere in this codebase. This comment probably needs to be updated

packages/browser/src/openfeature/devtools-provider.ts
sameerank, line 1: Would be a good idea to create `packages/browser/test/openfeature/devtools-provider.ts‎` and write some tests for the override behavior

packages/browser/src/openfeature/devtools-provider.ts
```diff
@@ -0,0 +1,99 @@
...
+  try {
+    return JSON.parse(localStorage.getItem(OVERRIDES_KEY) ?? '{}') as Record<string, DDFlagOverride>
+  } catch {
+    return {}
+  }
+}
+
+function resolveOverride<T>(flagKey: string, expectedTypes: DDFlagOverrideType[]): ResolutionDetails<T> {
+  const override = readOverrides()[flagKey]
+  if (!override || !expectedTypes.includes(override.type)) {
```
sameerank, line 41: What happens here if there's a value-type mismatch? E.g. `{ type: 'BOOLEAN', value: 'yes' }`

sameerank (commented):
Design looks good at a high level! Mainly would like to see some tests to help document the override and fall through paths

packages/browser/src/init.ts
kellyw1806: yes, ill remove the orgContext/userContext and just have generic context maps -> defaults to user so existing callers aren't broken

packages/browser/src/openfeature/devtools-provider.ts
kellyw1806: the closest existing type is VariantType but that is in node-server and browser cannot import from that sibling package. we could move this type up to flagging-core?

packages/browser/src/openfeature/devtools-provider.ts
kellyw1806: i think indexedDB is async-only (no synchronous read), whereas our provider currently reads overrides synchronously at flag-eval time. this means that switching to indexedDB would make every flag eval async which will have a larger blast radius (callers would need to await and you couldn't read flags inline during render). so for that reason, i thinkg LocalStorage might fit better. also indexeddb has a more complex API and its main strengths like the large capacity, indexing aren't really things we use here since this is just a tiny key value map. we could revisit indexeddb if we think that the data will grow?

leoromanovsky: I'm sorry to not see that on the first PR review. Loading from local storage is (relatively) slow and JSON parsing during flag evaluation adds overhead. Basically every single flag evaluation will pay the same penalty, but basically be loading the same data. Could your provider instead load from local storage (or indexdb) once on creation and re-use it?

PR #320, opened 2026-06-29
fix(browser): add descriptive reason to aborts from superseded context

Wittiest:
## Motivation

The provider calls AbortController.abort() with no argument when a newer setContext supersedes an in-flight precompute-assignments fetch. The browser then surfaces the generic "signal is aborted without reason" AbortError, which RUM/error trackers record as noise even though the provider handles the abort cleanly and flag resolution is unaffected.

## Changes

Pass a named AbortError with a descriptive reason so the rejection is self-documenting and consumers can reliably filter intentional supersession aborts without masking genuine fetch failures. The DOMException is still named 'AbortError', so all existing signal.aborted handling is preserved.

## Checklist

- [ ] Updated Documentation
- [ ] Tested locally
- [ ] Tested on staging

### DIFF
--- a/packages/browser/src/openfeature/provider.ts
+++ b/packages/browser/src/openfeature/provider.ts
@@ -145,7 +145,9 @@ export class DatadogProvider implements Provider {
     }
 
     // abort any previous setContext operation
-    this.contextUpdateAbortController.abort()
+    this.contextUpdateAbortController.abort(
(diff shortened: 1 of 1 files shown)


sameerank (approved):
Thank you!

I've staged a follow up PR with a test that passed for me locally to verify this change https://github.com/DataDog/openfeature-js-client/pull/321/changes

PR #321, opened 2026-06-29
test(browser): assert abort reason is descriptive DOMException on superseded context

sameerank:
## Motivation

Companion test for #320.

PR #320 passes a named `DOMException` as the abort reason when a newer `setContext` supersedes an in-flight fetch, so that RUM/error trackers can filter intentional aborts. This PR adds a unit test that verifies that contract.

## Changes

Adds one test to the `concurrent request ordering` block in `provider.spec.ts` that:
1. Starts a context update and captures the `AbortSignal` passed to the fetch
2. Fires a second context update to trigger the supersession
3. Asserts `signal.aborted === true`, `signal.reason instanceof DOMException`, `name === 'AbortError'`, and the expected message


leoromanovsky (approved):
TY!

PR #324, opened 2026-06-30
feat(browser): add DatadogDevtools wrapper provider for flag overrides

kellyw1806:
## Context

The Datadog Configurator currently applies flag overrides through wrapper functions in web-ui that short-circuit before OpenFeature runs. This means hooks never fire for overridden flags, there's no single observation point to record what flags a page evaluated, and customers using the SDK directly have no override mechanism at all.

This PR moves overrides inside the OpenFeature chain by introducing `DatadogDevtools` — a provider wrapper that sits in front of any inner provider and intercepts evaluations where an override exists. The core `DatadogProvider` is untouched.

This is part of the Browser Extension Project: [RFC](https://docs.google.com/document/d/1bEZqKgT_sX2gVVdq8bR8BlelkcFzrcsocNfO0Lj9_38/edit?pli=1&tab=t.0)

## What changed

**`DatadogDevtools`** — a new provider wrapper exported from `@datadog/openfeature-browser`:
- Wraps any OpenFeature provider — no `MultiProvider` dependency required
- Loads overrides **once on `initialize()`** from `localStorage['dd.dd_flag.overrides']` into memory — zero localStorage reads per flag evaluation
- On override hit: returns the value with `flagMetadata.overridden = true` and `reason: STATIC`, does not call the inner provider
- On override miss: delegates directly to the inner provider unchanged
- Supports all 5 flag types: `BOOLEAN`, `STRING`, `INTEGER`, `NUMERIC`, `JSON`
- Runtime type validation — throws `TypeMismatchError` and `console.warn`s if the declared type and actual value type don't match, so bad overrides are visible rather than silently falling through to the default
- Writes an enablement marker (`dd.dd_flag.devtools = "enabled"`) on `initialize()` so the browser extension can detect devtools are active; removes it on `onClose()`
- Forwards `hooks`, `events`, `status`, `onContextChange`, and `onClose` to the inner provider

**Tests** — 31 tests covering: all 5 override types, falsy values (`false`, `0`), miss/delegation for all 4 resolve methods, cross-type delegation, `TypeMismatchError` + `console.warn` for all mismatch cases, `null` JSON value rejection, overrides loaded exactly once (localStorage read count asserted), marker written/removed, lifecycle methods delegated, inner provider errors bubbling through, malformed localStorage.

## How customers use this

**Production — no change needed:**
```ts
import { DatadogProvider } from '@datadog/openfeature-browser'

await OpenFeature.setProvider(new DatadogProvider(config))
```

**Development — wrap the provider:**
```ts
import { DatadogProvider, DatadogDevtools } from '@datadog/openfeature-browser'

await OpenFeature.setProvider(new DatadogDevtools(new DatadogProvider(config)))
```

No `MultiProvider`, no extra dependencies. When dev tools are not needed, `DatadogDevtools` is absent from the bundle entirely.

Overrides live in `localStorage['dd.dd_flag.overrides']` — written by the browser extension in practice, but settable manually for testing:

```js
localStorage.setItem('dd.dd_flag.overrides', JSON.stringify({
  'my-boolean-flag': { type: 'BOOLEAN', value: true },
  'app-theme':        { type: 'STRING',  value: 'dark' },
  'rollout-percent':  { type: 'NUMERIC', value: 100 },
}))
```

Overrides persist across reloads. Within a session they're loaded once at startup — mid-session `localStorage` changes take effect on the next reload. Overridden evaluations return `flagMetadata.overridden === true`.

## Test plan
- [ ] Override hit: set `dd.dd_flag.overrides`, evaluate flag — resolves to override, inner provider not called
- [ ] Override miss: no entry — delegates to inner provider normally
- [ ] Type mismatch: `{ type: 'BOOLEAN', value: 'yes' }` — `console.warn` fires, falls back to default
- [ ] `null` JSON value: `{ type: 'JSON', value: null }` — rejected, `console.warn` fires
- [ ] Falsy values: `{ type: 'BOOLEAN', value: false }` and `{ type: 'NUMERIC', value: 0 }` — returned correctly
- [ ] Enablement marker: `localStorage.getItem('dd.dd_flag.devtools') === 'enabled'` after init
- [ ] `yarn workspace @datadog/openfeature-browser test` — all 31 tests pass

### DIFF
--- a/packages/browser/test/openfeature/devtools-provider.spec.ts
+++ b/packages/browser/test/openfeature/devtools-provider.spec.ts
@@ -1,0 +1,397 @@
+import type { EvaluationContext, JsonValue, Logger, Provider, ResolutionDetails } from '@openfeature/web-sdk'
+import { OpenFeatureEventEmitter, TypeMismatchError } from '@openfeature/web-sdk'
+import { DatadogDevtools } from '../../src/openfeature/devtools-provider'
+
+const OVERRIDES_KEY = 'dd.dd_flag.overrides'
+const DEVTOOLS_MARKER_KEY = 'dd.dd_flag.devtools'
+
+const noopLogger = {} as Logger
+const emptyContext: EvaluationContext = {}
+
+function makeInner(overrides: Partial<Provider> = {}): jest.Mocked<Provider> {
+  return {
+    metadata: { name: 'mock-inner' },
+    runsOn: 'client',
+    hooks: [],
+    events: new OpenFeatureEventEmitter(),
+    initialize: jest.fn().mockResolvedValue(undefined),
+    onClose: jest.fn().mockResolvedValue(undefined),
+    onContextChange: jest.fn().mockResolvedValue(undefined),
+    resolveBooleanEvaluation: jest.fn().mockReturnValue({ value: false, reason: 'DEFAULT' }),
+    resolveStringEvaluation: jest.fn().mockReturnValue({ value: 'inner-default', reason: 'DEFAULT' }),
+    resolveNumberEvaluation: jest.fn().mockReturnValue({ value: 0, reason: 'DEFAULT' }),
+    resolveObjectEvaluation: jest.fn().mockReturnValue({ value: {}, reason: 'DEFAULT' }),
+    ...overrides,
+  } as jest.Mocked<Provider>
+}
+
+function setOverrides(overrides: Record<string, unknown>) {
+  localStorage.setItem(OVERRIDES_KEY, JSON.stringify(overrides))
+}
+
+describe('DatadogDevtools', () => {
+  beforeEach(() => {
+    localStorage.clear()
+  })
+
+  describe('initialization', () => {
+    it('loads overrides from localStorage once on initialize', async () => {
+      setOverrides({ 'my-flag': { type: 'BOOLEAN', value: true } })
+      const inner = makeInner()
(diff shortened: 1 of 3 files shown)


packages/browser/test/openfeature/devtools-provider.spec.ts
```diff
@@ -0,0 +1,415 @@
...
+  localStorage.setItem(OVERRIDES_KEY, JSON.stringify(overrides))
+}
+
+describe('DatadogDevtools', () => {
+  beforeEach(() => {
+    localStorage.clear()
+  })
+
+  describe('initialization', () => {
+    it('loads overrides from localStorage once on initialize', async () => {
```
leoromanovsky, line 38: 👍

packages/browser/test/openfeature/devtools-provider.spec.ts
```diff
@@ -0,0 +1,415 @@
...
+        throw new Error('storage unavailable')
+      })
+
+      const wrapper = new DatadogDevtools(makeInner())
+      await expect(wrapper.initialize(emptyContext)).resolves.toBeUndefined()
+
+      jest.restoreAllMocks()
+    })
+
+    it('falls back to empty overrides when localStorage contains malformed JSON', async () => {
```
leoromanovsky, line 89: very nice

leoromanovsky (approved):

PR #331, opened 2026-07-08
fix(core): serialize precomputed.response in configurationToString

btthomas:
## Motivation

`configurationToString` does not round-trip: serializing a `FlagsConfiguration` and parsing it back with `configurationFromString` loses all flags. Found while evaluating reuse of `@datadog/flagging-core`'s wire helpers from the `dd-sdk-reactnative` SDK.

## Changes

`configurationToString` stringified the **entire** `precomputed` object into the wire's `precomputed.response` field instead of just `precomputed.response`:

```ts
// before
response: JSON.stringify(configuration.precomputed),
// after
response: JSON.stringify(configuration.precomputed.response),
```

Because of this, `configurationFromString` (which parses `wire.precomputed.response` back into `response`) nested the real response one level too deep — `precomputed.response` became `{ response, context, fetchedAt }`, so `precomputed.response.data` was `undefined` and every flag was lost on a serialize→parse cycle.

The CDN-produced wire is unaffected (it stringifies `response` correctly); the bug only manifests when `configurationToString` produces the wire (persisting or re-serializing a config).

The fix mirrors `configurationFromString`. Adds `wire.test.ts` (the file's first test) covering:
- round-trip identity
- flags still readable after a round-trip
- the serialized `precomputed.response` shape (regression guard)
- version and malformed-input fallbacks

Verified the new test fails against the old code (3 targeted failures) and passes with the fix.

## Test instructions

`yarn --cwd packages/core test` — 5 suites / 33 tests pass. `yarn lint` and `yarn --cwd packages/core typecheck` are clean.

## Checklist

- [ ] Updated Documentation
- [x] Tested locally
- [ ] Tested on staging

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/packages/core/src/configuration/wire.test.ts
+++ b/packages/core/src/configuration/wire.test.ts
@@ -1,0 +1,59 @@
+import type { FlagsConfiguration } from './configuration'
+import { configurationFromString, configurationToString } from './wire'
+
+const configuration: FlagsConfiguration = {
+  precomputed: {
+    response: {
+      data: {
+        attributes: {
+          createdAt: '2026-07-06T23:01:56.822Z',
+          flags: {
+            'my-flag': {
+              allocationKey: 'alloc-1',
+              variationKey: 'true',
+              variationType: 'boolean',
+              variationValue: true,
+              reason: 'STATIC',
+              doLog: true,
+              extraLogging: {},
+            },
+          },
+        },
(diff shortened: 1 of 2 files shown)


sameerank (approved):
Nice thanks for catching this!

PR #336, opened 2026-07-10
FFL-2753 Add browser DatadogOfflineProvider

sameerank:
## Summary

Add an opt-in browser `DatadogOfflineProvider` that evaluates a caller-supplied `FlagsConfiguration` without fetching or polling configuration.

This PR is stacked on [#344](https://github.com/DataDog/openfeature-js-client/pull/344), which provides the portable wire decoder, protobuf-backed rules configuration, and capability-specific browser entry points.

- add an internal `DatadogCoreProvider` base shared by `DatadogProvider` and `DatadogOfflineProvider`
- export `DatadogOfflineProvider` from the full browser entry point and the protobuf-free `/precomputed` entry point
- load offline configuration through `setConfiguration()`, matching the React Native provider API
- keep the shared core provider internal to the browser package
- move the shared top-level `evaluate(...)` orchestration into `@datadog/flagging-core`
- use one core capability selector for evaluation and provider lifecycle decisions
- evaluate a matching precomputed branch first, then fall back to valid rules
- keep capability errors independent: an invalid sibling does not block a valid capability
- distinguish never configured from configured but unusable
- return and emit `PARSE_ERROR` for known configuration parsing failures, reserving `GENERAL` for unexpected failures
- emit standard OpenFeature error-event messages from online and offline providers, with recognized OpenFeature `errorCode` values preserved as a runtime extension
- reject precomputed-only configuration when its stored context does not match
- keep the existing `DatadogProvider` API and no-configuration fallback behavior

## Why

[FFL-2753](https://datadoghq.atlassian.net/browse/FFL-2753) adds the browser provider layer for portable configuration and dynamic context. Customers can load a `FlagsConfiguration`, including one decoded from ConfigurationWire, and evaluate locally. Rules configurations handle OpenFeature context changes without a network round trip.

The provider hierarchy and configuration API follow the same shape as the React Native integration: an internal base owns the common OpenFeature evaluation surface, while consumers use concrete online and offline providers. Caller-supplied configuration belongs to the offline provider and is loaded with `setConfiguration()` rather than passed to its constructor.

## Capability precedence

1. Use valid precomputed data when its context matches.
2. Otherwise, evaluate valid rules.
3. If no valid capability can serve the context and a capability or envelope failed parsing, return `PARSE_ERROR`.
4. Otherwise, return `INVALID_CONTEXT` for mismatching precomputed data.
5. Return `PROVIDER_NOT_READY` only when configuration has never been supplied.

This keeps valid and invalid configuration branches independent and prevents browser lifecycle state from disagreeing with evaluation results. Regression tests cover the complete precomputed/rules validity matrix, malformed envelopes, explicitly empty configurations, and the never-configured state.

## Dependency and merge order

1. Merge #344 into `main`.
2. Retarget this PR to `main`.
3. Merge this PR.

The default browser entry point supports both rules and precomputed configuration. Precomputed-only consumers can import `DatadogOfflineProvider` and `configurationFromString` from `@datadog/openfeature-browser/precomputed`; the entry-point smoke test verifies that path loads no Protobuf-ES modules.

## Behavior

- construct `DatadogOfflineProvider`, then load or replace configuration with `setConfiguration()`
- `initialize()` validates the loaded configuration and context
- `onContextChange()` validates the new context synchronously and lets OpenFeature handle lifecycle errors
- `setConfiguration()` emits `ConfigurationChanged`; recovery also emits `Ready`, while unusable configuration emits `Error`
- malformed configuration maps to OpenFeature `ParseError` and evaluation `PARSE_ERROR`
- provider error events contain `message` instead of the private `{ error }` shape; OpenFeature errors also carry `errorCode`
- precomputed configurations are used only when their stored context matches the evaluation context
- rules are used when precomputed data mismatches or the precomputed capability is invalid
- the existing `DatadogProvider` API and default fallback behavior remain unchanged

## Validation

- full unit suite passed: 429 core, 191 browser, and 55 node-server tests
- Biome checks passed
- public entry-point checks confirm the internal core provider is not exported
- precomputed CommonJS smoke check loaded zero Protobuf-ES modules

[FFL-2753]: https://datadoghq.atlassian.net/browse/FFL-2753


packages/browser/src/openfeature/core-provider.ts
```diff
@@ -0,0 +1,164 @@
...
+
+  status: ProviderStatus
+
+  private flagsConfiguration: FlagsConfiguration
+  private context: EvaluationContext = {}
+
+  constructor(options: CoreProviderOptions) {
+    this.events = new OpenFeatureEventEmitter()
+    this.flagsConfiguration = options.configuration
+    this.status = hasEvaluatableConfiguration(options.configuration) ? ProviderStatus.READY : ProviderStatus.ERROR
```
dd-oleksii: `status` is deprecated / unused, so we can leave it out
```suggestion
```

packages/browser/src/openfeature/core-provider.ts
```diff
@@ -0,0 +1,164 @@
...
+    this.context = context
+    const error = getConfigurationError(this.flagsConfiguration, this.context)
+    if (error) {
+      this.status = ProviderStatus.ERROR
+      throw error
+    }
+    this.status = ProviderStatus.READY
+  }
+
+  async onContextChange(_oldContext: EvaluationContext, newContext: EvaluationContext): Promise<void> {
```
dd-oleksii: minor: does it make sense to make `initialize` and `onContextChange` sync?
```suggestion
  onContextChange(_oldContext: EvaluationContext, newContext: EvaluationContext): Promise<void> {
```

packages/browser/src/openfeature/core-provider.ts
```diff
@@ -0,0 +1,164 @@
...
+    this.context = newContext
+    const error = getConfigurationError(this.flagsConfiguration, this.context)
+    if (error) {
+      this.status = ProviderStatus.ERROR
+      this.events.emit(ProviderEvents.Error, { error })
```
dd-oleksii: Why do we emit an event here but not on ready path?

packages/browser/src/openfeature/core-provider.ts
```diff
@@ -0,0 +1,164 @@
...
+    context: EvaluationContext,
+    logger: Logger
+  ): ResolutionDetails<FlagTypeToValue<T>> {
+    if (!hasEvaluatableConfiguration(this.flagsConfiguration)) {
+      return {
+        value: defaultValue,
+        reason: 'ERROR',
+        errorCode: 'PROVIDER_NOT_READY' as ErrorCode,
+      }
+    }
```
dd-oleksii: nitpick: this check looks like something that `evaluate` should handle

packages/browser/src/openfeature/core-provider.ts
```diff
@@ -0,0 +1,164 @@
...
+
+    return evaluate(this.flagsConfiguration, type, flagKey, defaultValue, this.getEvaluationContext(context), logger)
+  }
+
+  private getEvaluationContext(context: EvaluationContext): EvaluationContext {
+    if (Object.keys(context).length > 0 || Object.keys(this.context).length === 0) {
+      return context
+    }
+    return this.context
+  }
```
dd-oleksii: why is this needed? have you seen openfeature sdk passing the wrong context?

packages/browser/test/openfeature/provider.spec.ts
```diff
@@ -298,6 +298,7 @@ describe('DatadogProvider', () => {
         flagMetadata: {
           allocationKey: 'allocation-123',
           doLog: true,
+          extraLogging: { experiment: true },
```
dd-oleksii: major:
1. `extraLogging` is deprecated.
2. it's `Record<string, string>` (boolean not allowed).
3. `flagMetadata` only [allows](https://openfeature.dev/specification/types/#flag-metadata) `boolean`, `string`, and `number` values (nested object is not allowed)

packages/browser/README.md
```diff
@@ -90,6 +90,27 @@ console.log(result.value) // Flag value
...
+
+`CoreProvider` is an opt-in evaluation-only provider for applications that supply their own flags configuration, such as an SSR bootstrap or offline init payload. It does not fetch or poll configuration.
+
+For dynamic context, the generic configuration wire should contain rules-based flag configuration. Precomputed configuration can also be evaluated, but only for the matching context it was generated for.
+
+```javascript
+import { configurationFromString, CoreProvider } from '@datadog/openfeature-browser'
+import { OpenFeature } from '@openfeature/web-sdk'
+
+const configuration = configurationFromString(window.__DD_FLAGS_CONFIGURATION__)
```
dd-oleksii: minor: users might get confused about `__DD_FLAGS_CONFIGURATION__` — what's that and how it is set. Maybe we should hint that it's just a string and they should transfer it however they want, or say that `__DD_FLAGS_CONFIGURATION__` is set by SSR process?

```suggestion
const configuration = configurationFromString("...flags configuration string...")
```

packages/browser/README.md
```diff
@@ -90,6 +90,27 @@ console.log(result.value) // Flag value
...
+const provider = new CoreProvider({ configuration })
+
+await OpenFeature.setProviderAndWait(provider)
+
+await OpenFeature.setContext({ targetingKey: 'user-123', plan: 'enterprise' })
+
+const client = OpenFeature.getClient()
+const enabled = await client.getBooleanValue('new-checkout', false)
```
dd-oleksii: minor: web evalulation methods are sync:
```suggestion
const enabled = client.getBooleanValue('new-checkout', false)
```

packages/core/src/evaluation/evaluateForSubject.ts
```diff
@@ -96,6 +96,7 @@ export function evaluateForSubject<T extends FlagValueType>(
             allocationKey: allocation.key,
             variationType: variantTypeToFlagValueType(flag.variationType),
             doLog: !!allocation.doLog,
+            extraLogging: selectedSplit.extraLogging,
```
dd-oleksii: major: `extraLogging` is deprecated and violates `flagMetadata` spec requirements (only `string`, `boolean`, and `number` values are allowed — not nested object)

packages/browser/src/evaluation.ts
```diff
@@ -4,19 +4,47 @@ import type {
...
+import type { ErrorCode, EvaluationContext, FlagValueType, Logger, ResolutionDetails } from '@openfeature/web-sdk'
+
+const NOOP_LOGGER: Logger = {
+  debug: () => {},
+  info: () => {},
+  warn: () => {},
+  error: () => {},
+}
 
 export function evaluate<T extends FlagValueType>(
```
dd-oleksii: minor: are we able to move evaluation function to the core package?

packages/browser/src/openfeature/core-provider.ts
sameerank: Good call for `onContextChange`. The web provider signature allows sync context changes:

`onContextChange?(oldContext: EvaluationContext, newContext: EvaluationContext): Promise<void> | void;`
* https://github.com/open-feature/js-sdk/blob/bc470dca88afa26462b2513e2d106df324e7ba46/packages/web/src/provider/provider.ts#L34

So I can remove `async` there and change the return type to `void`, which also avoids OpenFeature treating rules-based context changes as async reconciliation.

For `initialize` though, I believe we still need to return a promise to match the shared provider contract:

`initialize?(context?: EvaluationContext): Promise<void>;`
* https://github.com/open-feature/js-sdk/blob/bc470dca88afa26462b2513e2d106df324e7ba46/packages/shared/src/provider/provider.ts#L128

The SDK calls `.then(...)` on the returned value to mark the provider ready, so a truly sync `initialize(): void` would not be compatible with the current OpenFeature implementation.
* https://github.com/open-feature/js-sdk/blob/bc470dca88afa26462b2513e2d106df324e7ba46/packages/shared/src/open-feature.ts#L256-L258

packages/browser/src/openfeature/core-provider.ts
sameerank: https://github.com/DataDog/openfeature-js-client/pull/336/changes/efb2034faf809fe6a22a1f28845138f3849a3a28 removes the `CoreProvider` status field

packages/browser/src/openfeature/core-provider.ts
sameerank: https://github.com/DataDog/openfeature-js-client/pull/336/commits/9cf1cccc419bb2e189c335b1dad3303c75108cfa makes `CoreProvider` context changes synchronous

packages/browser/src/openfeature/core-provider.ts
sameerank: https://github.com/DataDog/openfeature-js-client/pull/336/commits/f151909fdbb7f81e11815520066e5392c7a98878 removes the explicit error emit from `onContextChange`, so `CoreProvider` treats OpenFeature lifecycle methods consistently:
* `initialize(...)` updates context and throws if invalid; OpenFeature handles provider readiness/error lifecycle.
* `onContextChange(...)` updates context and throws if invalid; OpenFeature handles the context-change error lifecycle.
* `setConfiguration(...)` still emits Ready, ConfigurationChanged, or Error because that method is our own provider API, not an OpenFeature lifecycle callback.

packages/browser/src/openfeature/core-provider.ts
sameerank: https://github.com/DataDog/openfeature-js-client/pull/336/commits/f011597aaf12ba59b42ae565290d46592b5b1e66 moves missing configuration fallback into `evaluate`

packages/browser/src/openfeature/core-provider.ts
sameerank: Yep, unnecessary. https://github.com/DataDog/openfeature-js-client/pull/336/commits/21195fb847e662e92572c71a616ab621ba4e7fe9 removes `getEvaluationContext` and relies on the OpenFeature to pass the right evaluation context into the provider resolve methods

packages/browser/test/openfeature/provider.spec.ts
sameerank: https://github.com/DataDog/openfeature-js-client/pull/336/commits/1ad76ed5f1e86c46e68930be20fcc656feda23a7 removes `extraLogging` from flag metadata

packages/browser/README.md
sameerank: https://github.com/DataDog/openfeature-js-client/pull/336/commits/34b3e808975fb3b8de471d88be43b4257bfe3ae0 updates the portable configuration string example

packages/browser/README.md
sameerank: https://github.com/DataDog/openfeature-js-client/pull/336/commits/9371b90d2ab9a246292422639eb0d5be7e370317 documents browser evaluations as synchronous

packages/browser/src/evaluation.ts
sameerank: Sure, moved in https://github.com/DataDog/openfeature-js-client/pull/336/commits/86081daa5d47f433bbbb990e3352cfc86a38858e

Will have to do a 2.0.1 release for `@datadog/flagging-core` and update the `@datadog/openfeature-browser` dependency before we can release

packages/browser/src/openfeature/core-provider.ts
sameerank: A follow up https://github.com/DataDog/openfeature-js-client/pull/336/commits/9c4e4154f1869bb38dde7cd6849cc5374f3f35ba fixes the consequence of this change: `CoreProvider` should return `PROVIDER_NOT_READY` when it has no configuration, but the pre-existing `DatadogProvider` tests expected the old behavior: return the caller’s default value with reason `DEFAULT` before fetched config is available. Or should I switch them both to `PROVIDER_NOT_READY`? I hesitate because changing the server side evaluation feels like creeping beyond the scope of what we're trying to do here

packages/browser/README.md
```diff
@@ -131,6 +131,28 @@ names, but ignores rules-based entries:
 import { configurationFromString, DatadogProvider } from '@datadog/openfeature-browser/precomputed'
 ```
 
+### Using DatadogOfflineProvider with portable configuration
```
leoromanovsky, line 157: 👍

packages/core/src/evaluation/evaluation.ts
```diff
@@ -9,6 +16,104 @@ import { createEvaluationTimestampMetadata } from './evaluationMetadata'
...
+    return evaluateRulesBasedConfiguration(
+      selection.configuration.response,
+      type,
+      flagKey,
+      defaultValue,
+      context,
+      logger
+    )
+  }
+
```
leoromanovsky, line 50: else? warning?

packages/core/test/evaluation/evaluateForSubject.spec.ts
```diff
@@ -151,6 +151,7 @@ describe('evaluateForSubject', () => {
               {
                 variationKey: 'treatment',
                 serialId,
+                extraLogging: { campaign: 'summer' },
```
leoromanovsky: Unused or am misunderstanding this test?

packages/core/src/evaluation/evaluation.ts
```diff
@@ -71,6 +176,70 @@ export function evaluateRulesBasedConfiguration<T extends FlagValueType>(
...
+  type: T,
+  flagKey: string,
+  defaultValue: FlagTypeToValue<T>
+): ResolutionDetails<FlagTypeToValue<T>> {
+  const flagError = precomputed.flagErrors ? getOwnProperty(precomputed.flagErrors, flagKey) : undefined
+  if (flagError) {
+    return {
+      value: defaultValue,
+      reason: 'ERROR',
+      errorCode: 'PARSE_ERROR' as ErrorCode,
```
leoromanovsky, line 199: We don't have an error code that the server can provide?

packages/browser/test/openfeature/offline-provider.spec.ts
```diff
@@ -0,0 +1,258 @@
...
+      message: 'Flags configuration contains no usable capability',
+      errorCode: 'PARSE_ERROR',
+    })
+    expect(provider.resolveBooleanEvaluation('missing-flag', true, {}, logger)).toMatchObject({
+      errorCode: 'PARSE_ERROR',
+      errorMessage: 'Flags configuration contains no usable capability',
+    })
+  })
+
+  it('uses parse errors for malformed configured capabilities', async () => {
```
leoromanovsky, line 257: very thorough in this spec file thank you.

leoromanovsky (approved):
Do you see value in the offline provider "looping over" the ffe system test data json fixtures like the "live" provider does?

packages/core/src/evaluation/evaluation.ts
sameerank: Not currently. The precompute implementation omits flags whose UFC evaluation fails; the serialized assignment has no error-code field. This `flagError` is generated locally when the SDK receives but cannot parse a flag entry, so it maps to OpenFeature `PARSE_ERROR`.

packages/core/src/evaluation/evaluation.ts
sameerank: At this point, the only possible state is `error`. We will already structure an OpenFeature error details; and additionally warning on every evaluation could get noisy

packages/core/test/evaluation/evaluateForSubject.spec.ts
sameerank: `extraLogging` is specifically a deprecated field https://github.com/DataDog/openfeature-js-client/pull/344#discussion_r3654761302

And so this test reinforces that we don't copy it into OpenFeature flag metadata. But now I'm thinking this regression assertion is not important in this PR, so I'm going to remove it https://github.com/DataDog/openfeature-js-client/pull/336/commits/dde93ea4d1352a4d004bcb38de5d90a969c95e48

packages/browser/src/openfeature/provider.ts
```diff
@@ -280,51 +278,20 @@ export class DatadogProvider implements Provider {
...
     flagKey: string,
-    defaultValue: number,
+    defaultValue: FlagTypeToValue<T>,
     _context: EvaluationContext,
     _logger: Logger
-  ): ResolutionDetails<number> {
-    return evaluate(this.flagsConfiguration, 'number', flagKey, defaultValue, this.evaluationContext)
-  }
+  ): ResolutionDetails<FlagTypeToValue<T>> {
+    if (!this.flagsConfiguration.precomputed && !this.flagsConfiguration.rules) {
```
btthomas: > The early return for configurations with no precomputed or rules branch masks every parser error. For a configuration such as { rulesError: "malformed rules" }, the online provider returns reason DEFAULT, while the shared evaluator and offline provider return PARSE_ERROR. This contradicts the pull request's independent-error behavior and regresses the previous precomputed-error handling.

`reason: 'DEFAULT'` should be just for when the whole flagsConfiguration object is empty, right? If there is other stuff swimming around,  perhaps we should return `'PARSE_ERROR'` to match the standard evaluator.

packages/browser/src/openfeature/offline-provider.ts
```diff
@@ -0,0 +1,82 @@
...
+
+  constructor() {
+    super()
+  }
+
+  getConfiguration(): FlagsConfiguration | undefined {
+    return this.flagsConfiguration
+  }
+
+  setConfiguration(configuration: FlagsConfiguration): void {
```
btthomas, line 30: > The state in which no OpenFeature context has been supplied is represented as {}. Loading a valid context-specific precomputed configuration before provider registration therefore emits INVALID_CONTEXT, even when the caller subsequently registers the provider with the exact embedded context. The provider cannot distinguish an unknown context from an explicitly selected empty context.

Some customers are just going to use flags as kill switches without any context. Can we track whether init was called with a context separately from the current context value? We could add a test that calls `setConfiguration()` with a precomputed config w/ context and then call setProviderAndWait with that context, right?

btthomas (approved):
Some AI generated reviews and my comments on them. Very close!

packages/browser/src/openfeature/core-provider.ts
dd-oleksii: We can still have completely sync method, returning `Promise.resolve()`

packages/browser/src/openfeature/core-provider.ts
dd-oleksii: We want consistent behavior and the provider should return `PROVIDER_NOT_READY` if it doesn't have a configuration. I think it's fine to do this in this PR but we can also add another PR to the stack if you want to segregate the fix

packages/core/src/evaluation/evaluateForSubject.ts
sameerank: https://github.com/DataDog/openfeature-js-client/commit/d166c7f73d102fb3c159727deaf4656b619c9f06

packages/browser/src/openfeature/core-provider.ts
sameerank: Updated https://github.com/DataDog/openfeature-js-client/commit/184937fd8a93dd25b650570fb85fba2a2292c6d4

packages/browser/src/openfeature/provider.ts
sameerank: Fixed in https://github.com/DataDog/openfeature-js-client/commit/ddcb7701fc53c4f73889bd3aa8ea2f504726905d

The online provider now delegates these states to the shared evaluator. A provider that has never received configuration returns `PROVIDER_NOT_READY`, while a supplied malformed configuration such as `{ rulesError: "..." }` returns `PARSE_ERROR`

packages/browser/src/openfeature/offline-provider.ts
sameerank: Fixed in https://github.com/DataDog/openfeature-js-client/commit/9fd61c458d259eb51973203d9a006715851d19c5

The provider now represents “not initialized yet” as `undefined`, separately from an explicitly selected `{}` context. Therefore, `setConfiguration()` before registration stores the configuration without validating it against a placeholder context; initialization then validates it against the actual OpenFeature context.
Context-agnostic configurations, such as kill switches, remain valid with `{}`. I also added the suggested regression test covering `setConfiguration(precomputed)` followed by `setProviderAndWait` with the matching embedded context

btthomas (approved):

packages/browser/README.md
```diff
@@ -152,6 +152,28 @@ import {
 } from '@datadog/openfeature-browser/rules-based'
 ```
 
+### Using DatadogOfflineProvider with portable configuration
+
+`DatadogOfflineProvider` is an opt-in evaluation-only provider for applications that supply their own flags configuration, such as an SSR bootstrap or offline init payload. It does not fetch or poll configuration.
```
BenoitZugmeyer: **suggestion:** Difference between `DatadogOfflineProvider` and `DatadogProvider` is more than just fetching/polling configuration. It also skips RUM integration, flag evaluation and exposure tracking. Which... makes sense if we want true "offline" (= 0 access to Datadog). But documentation don't mention that. Is this intended?

packages/browser/src/openfeature/provider.ts
```diff
@@ -223,7 +221,7 @@ export class DatadogProvider implements Provider {
           } else {
             // Otherwise, this is a legitimate error
             this.status = ProviderStatus.ERROR
-            this.events.emit(ProviderEvents.Error, { error })
+            this.events.emit(ProviderEvents.Error, toProviderErrorEvent(error))
```
BenoitZugmeyer, line 224: **nitpick:** this looks like a small breaking change, no? Are we sure no user is using `event.error`?

packages/browser/README.md
sameerank: “Offline” here refers to configuration delivery, not the absence of telemetry.

[The RFC](https://docs.google.com/document/d/1OWNBtXtSk535VXqf-9fqsAmU9W8kpFLAwxYi2y1qyQQ/edit?tab=t.0) defines the evaluation-only provider as not fetching, polling, or managing context lifecycle, while exposure and evaluation logging are separate composable hooks. This PR establishes the provider/configuration lifecycle, and #349 adds optional tracking parity without making context changes trigger configuration fetches.

I’ll clarified the wording here so we don’t present “no telemetry” as part of the offline contract.

sameerank: Updated in https://github.com/DataDog/openfeature-js-client/pull/336/commits/fc10c2e65d4a93a07fc5177f342590b3b8b2ba29

packages/browser/src/openfeature/provider.ts
sameerank: Good catch! I restored the raw `error` property for compatibility while retaining the new `message` and `errorCode` fields. Also added regression coverage: https://github.com/DataDog/openfeature-js-client/pull/336/commits/fc10c2e65d4a93a07fc5177f342590b3b8b2ba29

packages/browser/src/openfeature/core-provider.ts
sameerank: This is now consistent: missing configuration returns `PROVIDER_NOT_READY` for both providers.

https://github.com/DataDog/openfeature-js-client/pull/336/commits/639699c1357d47da550188662ad6907678c704ca

The online-provider regression coverage was aligned here:

https://github.com/DataDog/openfeature-js-client/pull/336/commits/ddcb7701fc53c4f73889bd3aa8ea2f504726905d

PR #339, opened 2026-07-17
Release canonical evaluator patches

leoromanovsky:
## Motivation

PR #275 moved the canonical UFC evaluator behavior and fixture coverage into `@datadog/flagging-core`. The shared core package and both consumers need coordinated patch releases so Node.js and browser users receive the corrected evaluation reasons and malformed-configuration handling.

## Changes

- Bump `@datadog/flagging-core` from `2.0.0` to `2.0.1`.
- Bump `@datadog/openfeature-browser` from `1.2.3` to `1.2.4`.
- Bump `@datadog/openfeature-node-server` from `2.0.0` to `2.0.1`.
- Pin both consumers to the exact `@datadog/flagging-core@2.0.1` version and update the lockfile.
- Document the canonical evaluator fixes and shared fixture coverage in the changelog.

## Decisions

- Release all three affected packages because the implementation lives in core and internal package dependencies are intentionally exact.
- Keep independent package versions and tags rather than creating a unified version.
- After merge, publish in dependency order: core, browser, then Node.js.

## Validation

- `bash scripts/internal-deps-validate.sh`
- `bash scripts/versions-validate.sh`
- `yarn licenses:validate`
- `yarn lint`
- `yarn format:check`
- `BUILD_MODE=release yarn build`
- `yarn typecheck` (after the core declaration build)
- `yarn test` (457 tests across 22 suites)
- `yarn test:browser-install`
- `yarn test:node-install` (7/7 consumer checks)
- `yarn test:node-install:with-of` (7/7 consumer checks)
- `yarn lerna run pack --stream`

All validation ran with Node.js 24.16.0, matching the repository's Node 24 CI target.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,16 @@
 
 ---
 
+## @datadog/flagging-core@2.0.1, @datadog/openfeature-browser@1.2.4, @datadog/openfeature-node-server@2.0.1
+
+**Bug Fixes:**
+
+- Align shared UFC evaluation reasons and malformed-configuration isolation with the canonical cross-SDK fixtures ([#275](https://github.com/DataDog/openfeature-js-client/pull/275)) [BROWSER] [NODE-SERVER]
+
+**Internal Changes:**
+
+- Source evaluator regression coverage from `DataDog/ffe-system-test-data` ([#275](https://github.com/DataDog/openfeature-js-client/pull/275)) [BROWSER] [NODE-SERVER]
+
 ## @datadog/flagging-core@2.0.0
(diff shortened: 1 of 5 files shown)


typotter (approved):
LGTM++

We'll need to rip out the legacy lerna.json and related scripts at some point now that we've moved well past the synced versioning.

sameerank (approved):
If you create tags on this branch, please remember to use the merge and not squash so that lerna can still find the last commit with the tag

https://github.com/DataDog/openfeature-js-client/pull/315/changes

PR #344, opened 2026-07-24
[FFL-2835] Parse flags configuration wire format

sameerank:
## Summary

- add the opaque `FlagsConfigurationWire` contract and decode version 1 precomputed JSON and protobuf rules payloads in `@datadog/flagging-core`
- map the current UFC protobuf schema into the shared `FlagsConfiguration` model, including conditions, partition keys, shard modes, split reasons, and feature-level handling
- accept configurations that Protobuf-ES can parse and defer JavaScript-specific validation until the relevant evaluation path, returning scoped `PARSE_ERROR` details instead of dropping otherwise usable flags
- preserve whole-envelope, precomputed-capability, rules-capability, and flag-level parse errors independently, without blocking a valid sibling capability
- preserve the decoded protobuf message so rules configurations can be serialized back to the wire format, including unknown fields
- keep parse error messages deterministic and free of configuration-derived values so evaluation-event aggregation remains low-cardinality
- lazily compile regexes, compute partition hashes once per allocation, and validate condition ordering without per-evaluation cycle sets
- expose a full-featured `@datadog/openfeature-browser` entry point and an API-compatible, protobuf-free `@datadog/openfeature-browser/precomputed` capability entry point
- configure Protobuf-ES with internal fallbacks for React Native runtimes without native text encoding or `BigInt`
- serve as the base for the stacked CoreProvider follow-up in [#336](https://github.com/DataDog/openfeature-js-client/pull/336)

## Why

Browser and React Native SDK consumers need to turn the ConfigurationWire response into the shared configuration object used by local flag evaluation. Keeping the decoder in the core package makes the wire contract reusable. Capability-specific browser entry points keep the synchronous API while allowing precomputed-only consumers to exclude Protobuf-ES entirely.

Jira: https://datadoghq.atlassian.net/browse/FFL-2835

## Stacked follow-up

PR #336 is based on this branch and verifies that `CoreProvider` can evaluate the protobuf rules object returned by `configurationFromString`. Merge this PR first, then retarget #336 to `main`.

## Protobuf dependency trade-off

- **Benefits:** `@bufbuild/protobuf` provides a schema-driven, protobuf-conformant decoder, generated types, open-enum handling, and unknown-field round trips. This removes most hand-rolled protobuf logic and makes future canonical UFC schema updates safer and easier to review.
[...]

PR #346, opened 2026-07-28
[FFL-2871] Support static-context offline initialization

sameerank:
## Summary

This PR contains documentation and regression-test changes only. The browser provider implementation is introduced in [#336](https://github.com/DataDog/openfeature-js-client/pull/336), while the supported `getPrecomputedContext(configuration)` accessor comes from [#353](https://github.com/DataDog/openfeature-js-client/pull/353).

- document static offline initialization with a supported copy of the precomputed context
- use `getPrecomputedContext(configuration)` instead of inspecting the opaque wire or parsed configuration
- load configuration with `setConfiguration()` before registering the provider
- register the provider and matching context together with `OpenFeature.setProviderAndWait(provider, context)`
- treat `{}` and contexts containing only undefined values literally rather than adopting the embedded context
- preserve `INVALID_CONTEXT` behavior for context-specific precomputed snapshots that do not match
- add lifecycle coverage for initialization, context clearing, error transitions, and recovery

## Why

A context-specific precomputed configuration is a snapshot for the context used to generate it. The effective OpenFeature context must match so provider hooks and Datadog tracking observe the same subject used for evaluation.

The configuration wire and parsed configuration are opaque contracts. [#353](https://github.com/DataDog/openfeature-js-client/pull/353) provides `getPrecomputedContext(configuration)` so consumers can obtain a detached copy without inspecting internal fields.

An empty OpenFeature context is a real context. It does not mean “restore the embedded context.” Clearing context therefore supplies `{}` and puts a context-specific precomputed-only provider into `ERROR` until the matching context is set again.

## Behavior

- construct `DatadogOfflineProvider`, then load configuration with `setConfiguration()`
- no configuration fetch or polling is introduced
- matching precomputed configuration evaluates locally
- context-agnostic precomputed configuration continues to work with any context
- `{}`, `{ targetingKey: undefined }`, and cleared global context are treated literally
- mismatching context returns coded defaults with `INVALID_CONTEXT`
- setting the matching context recovers the provider to `READY`
- rules-based offline evaluation continues to support dynamic contexts without fetching
- the `/precomputed` entry point remains free of Protobuf-ES modules

## Dependency and merge order

This PR is stacked on [#336](https://github.com/DataDog/openfeature-js-client/pull/336). Its documented initialization flow also depends on `getPrecomputedContext` from [#353](https://github.com/DataDog/openfeature-js-client/pull/353).

1. Merge #344.
2. Merge #353 after retargeting it to `main`.
3. Retarget and merge #336.
4. Retarget this PR to `main` and merge it.
5. Retarget and merge #349 and #351 independently.

[PR #349](https://github.com/DataDog/openfeature-js-client/pull/349) adds opt-in exposure, evaluation, and RUM tracking parity to `DatadogOfflineProvider`. [PR #351](https://github.com/DataDog/openfeature-js-client/pull/351) adds explicit precomputed and rules-based configuration fetchers.

Jira: https://datadoghq.atlassian.net/browse/FFL-2871

## Validation

- full browser suite passed: 16 suites and 195 tests
- Biome and entry-point checks passed
- precomputed CommonJS smoke check loaded zero Protobuf-ES modules


leoromanovsky (approved):
👍

dd-oleksii (approved):

packages/browser/README.md
```diff
@@ -156,24 +156,29 @@ import {
...
+const context = getPrecomputedContext(configuration)
 
-await OpenFeature.setProviderAndWait(provider)
-
-await OpenFeature.setContext({ targetingKey: 'user-123', plan: 'enterprise' })
+if (context === undefined) {
+  await OpenFeature.setProviderAndWait(provider)
+} else {
+  await OpenFeature.setProviderAndWait(provider, context)
+}
```
BenoitZugmeyer, line 176: **thought:** When seeing this, I was surprised why it cannot be simplified as: 

```
await OpenFeature.setProviderAndWait(provider, getPrecomputedContext(configuration))
```

I understand now this is to work around OpenFeature typings, where `context` is not optional, but it is definitely surprising. Maybe a PR upstream could help simplifying this!

sameerank: Yep that’s why the conditional is needed. `getPrecomputedContext()` returns `EvaluationContext | undefined`, while the OpenFeature overload accepts either no second argument or a defined `EvaluationContext`.

We also cannot substitute `{}` for `undefined`, because this provider treats `{}` as an explicitly selected empty context. I agree an upstream typing/API change could make this simpler. The fix I'm imagining would live in the OpenFeature code though. Specifically [here](https://github.com/open-feature/js-sdk/blob/7959d6dd2c56aa3622f4abadc631c95b563d9209/packages/web/src/open-feature.ts#L83) and [here](https://github.com/open-feature/js-sdk/blob/7959d6dd2c56aa3622f4abadc631c95b563d9209/packages/web/src/open-feature.ts#L104) would let `context` be optional

```
setProviderAndWait(provider: Provider, context?: EvaluationContext): Promise<void>

setProviderAndWait(
  domain: string,
  provider: Provider,
  context?: EvaluationContext
): Promise<void>
```

PR #348, opened 2026-08-03
[FFL-2927] Bound browser assignment request latency

leoromanovsky:
## Motivation

Precomputed assignment fetches currently rely only on caller cancellation. A slow or unavailable edge can therefore delay provider initialization and evaluation-context changes for several seconds.

The browser SDK needs a bounded default request budget. Applications with different latency requirements must be able to configure that budget.

## Changes and Decisions

- Add `assignmentRequestTimeoutMs` and `assignmentRequestRetryCount` to the public provider configuration.
- Use a 1,000 ms timeout per attempt and one retry by default. The default budget is two attempts of up to one second each.
- Keep the timeout active through response body parsing, not only until response headers arrive.
- Retry timeouts, browser network errors, HTTP 408, and HTTP 5xx responses.
- Do not retry caller cancellation, HTTP 429, or other HTTP 4xx responses.
- Create a new abort controller for each attempt and preserve the caller signal across attempts.
- Define retry count as retries after the first attempt. Set it to zero to disable retries.
- Use the defaults when timeout or retry values are invalid.
- Preserve the existing failure and fallback path after the retry budget ends.

Jira: FFL-2927

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,12 @@
 
 ---
 
+## Unreleased
+
+**Features:**
+
+- Limit each browser assignment request to one second and retry transient failures once by default. Add timeout and retry-count configuration.
+
 ## @datadog/flagging-core@2.0.2, @datadog/openfeature-browser@1.2.5, @datadog/openfeature-node-server@2.0.2
 
 **Bug Fixes:**
--- a/packages/browser/README.md
+++ b/packages/browser/README.md
@@ -49,6 +49,10 @@ const provider = new DatadogProvider({
 
   // Enable flag evaluation tracking
   enableFlagEvaluationTracking: true,
+
+  // Assignment requests use one second per attempt and one retry by default
+  assignmentRequestTimeoutMs: 1000,
+  assignmentRequestRetryCount: 1,
 })
 ```
 
--- a/packages/browser/src/domain/configuration.ts
+++ b/packages/browser/src/domain/configuration.ts
@@ -71,6 +71,16 @@ export interface FlaggingInitConfiguration extends InitConfiguration {
    * Proxy URL for flagging configuration requests. If set, this will be used instead of the site parameter.
    */
   flaggingProxy?: string
+
+  /**
+   * Timeout for each flag assignment request in milliseconds (default: 1000ms).
+   */
+  assignmentRequestTimeoutMs?: number
+
+  /**
+   * Number of retries after a failed flag assignment request (default: 1).
+   */
+  assignmentRequestRetryCount?: number
 }
 
 export interface FlaggingConfiguration extends Configuration {
--- a/packages/browser/src/transport/fetchConfiguration.ts
+++ b/packages/browser/src/transport/fetchConfiguration.ts
@@ -1,4 +1,4 @@
-import type { FlagsConfiguration } from '@datadog/flagging-core'
+import type { FlagsConfiguration, PrecomputedConfigurationResponse } from '@datadog/flagging-core'
 import { timeStampNow } from '@datadog/js-core/time'
 import type { EvaluationContext } from '@openfeature/web-sdk'
 import type { FlaggingInitConfiguration } from '../domain/configuration'
@@ -15,6 +15,20 @@ type JSONAPIError = {
   }[]
 }
 
+const REQUEST_TIMEOUT_MS = 1_000
+const REQUEST_RETRY_COUNT = 1
+
+class HTTPResponseError extends Error {
+  constructor(
+    message: string,
+    readonly status: number
+  ) {
+    super(message)
+  }
+}
+
+class RequestTimeoutError extends Error {}
+
 async function getErrorMessage(response: Response) {
   if (
     response.headers.get('content-type') === 'application/vnd.api+json' ||
@@ -29,7 +43,77 @@ async function getErrorMessage(response: Response) {
   return response.statusText || 'Unknown error'
 }
 
+function isRetryableStatus(status: number) {
+  return status === 408 || status === 429 || (status >= 500 && status <= 599)
+}
+
+function isRetryableError(error: unknown) {
+  return (
+    error instanceof RequestTimeoutError ||
+    error instanceof TypeError ||
+    (error instanceof HTTPResponseError && isRetryableStatus(error.status))
+  )
+}
+
+async function fetchWithTimeout(
+  url: string,
+  init: RequestInit,
+  timeoutMs: number,
+  callerSignal?: AbortSignal
+): Promise<PrecomputedConfigurationResponse> {
+  const controller = new AbortController()
+  let timedOut = false
+  const abortFromCaller = () => controller.abort(callerSignal?.reason)
+
+  if (callerSignal?.aborted) {
+    abortFromCaller()
+  } else {
+    callerSignal?.addEventListener('abort', abortFromCaller, { once: true })
+  }
+
+  const timeout = setTimeout(() => {
+    timedOut = true
+    controller.abort()
+  }, timeoutMs)
+
+  try {
+    const response = await fetch(url, { ...init, signal: controller.signal })
+    if (!response.ok) {
+      let errorMessage: string
+      try {
+        errorMessage = await getErrorMessage(response)
+      } catch (error) {
+        if (isRetryableStatus(response.status)) {
+          const message = error instanceof Error ? error.message : 'Unknown error'
+          throw new HTTPResponseError(message, response.status)
+        }
+        throw error
+      }
+      throw new HTTPResponseError(`Failed to fetch flag configuration: ${errorMessage}`, response.status)
+    }
+    return (await response.json()) as PrecomputedConfigurationResponse
+  } catch (error) {
+    if (timedOut && !callerSignal?.aborted) {
+      throw new RequestTimeoutError('Flag configuration request timed out')
+    }
+    throw error
+  } finally {
+    clearTimeout(timeout)
+    callerSignal?.removeEventListener('abort', abortFromCaller)
+  }
+}
+
 export function createFlagsConfigurationFetcher(initConfiguration: FlaggingInitConfiguration) {
+  const configuredTimeout = initConfiguration.assignmentRequestTimeoutMs
+  const configuredRetryCount = initConfiguration.assignmentRequestRetryCount
+  const requestTimeoutMs =
+    typeof configuredTimeout === 'number' && Number.isFinite(configuredTimeout) && configuredTimeout > 0
+      ? configuredTimeout
+      : REQUEST_TIMEOUT_MS
+  const requestRetryCount =
+    typeof configuredRetryCount === 'number' && Number.isFinite(configuredRetryCount) && configuredRetryCount >= 0
+      ? Math.floor(configuredRetryCount)
+      : REQUEST_RETRY_COUNT
   let url: URL
   if (initConfiguration.flaggingProxy?.match('https?://')) {
     // If flaggingProxy has a protocol, use it as-is
@@ -66,10 +150,9 @@ export function createFlagsConfigurationFetcher(initConfiguration: FlaggingInitC
       stringifiedContext[key] = typeof value === 'string' ? value : JSON.stringify(value)
     }
 
-    const response = await fetch(url.toString(), {
+    const requestInit: RequestInit = {
       method: 'POST',
       headers: defaultHeaders,
-      signal,
       body: JSON.stringify({
         data: {
           type: 'precompute-assignments-request',
@@ -83,18 +166,25 @@ export function createFlagsConfigurationFetcher(initConfiguration: FlaggingInitC
           },
         },
       }),
-    })
-    if (!response.ok) {
-      const errorMessage = await getErrorMessage(response)
-      throw new Error(`Failed to fetch flag configuration: ${errorMessage}`)
     }
-    const precomputed = await response.json()
-    return {
-      precomputed: {
-        response: precomputed,
-        context,
-        fetchedAt: timeStampNow(),
-      },
+
+    for (let attempt = 0; attempt <= requestRetryCount; attempt += 1) {
+      try {
+        const precomputed = await fetchWithTimeout(url.toString(), requestInit, requestTimeoutMs, signal)
+        return {
+          precomputed: {
+            response: precomputed,
+            context,
+            fetchedAt: timeStampNow(),
+          },
+        }
+      } catch (error) {
+        if (signal?.aborted || attempt === requestRetryCount || !isRetryableError(error)) {
+          throw error
+        }
+      }
     }
+
+    throw new Error('Flag configuration request retry loop completed unexpectedly')
   }
 }
--- a/packages/browser/src/transport/fetchConfiguration.ts
+++ b/packages/browser/src/transport/fetchConfiguration.ts
@@ -15,8 +15,8 @@ type JSONAPIError = {
   }[]
 }
 
-const REQUEST_TIMEOUT_MS = 1_000
-const REQUEST_RETRY_COUNT = 1
+const DEFAULT_REQUEST_TIMEOUT_MS = 1_000
+const DEFAULT_REQUEST_RETRY_COUNT = 1
 
 class HTTPResponseError extends Error {
   constructor(
@@ -44,7 +44,7 @@ async function getErrorMessage(response: Response) {
 }
 
 function isRetryableStatus(status: number) {
-  return status === 408 || status === 429 || (status >= 500 && status <= 599)
+  return status === 408 || (status >= 500 && status <= 599)
 }
 
 function isRetryableError(error: unknown) {
@@ -109,11 +109,11 @@ export function createFlagsConfigurationFetcher(initConfiguration: FlaggingInitC
   const requestTimeoutMs =
     typeof configuredTimeout === 'number' && Number.isFinite(configuredTimeout) && configuredTimeout > 0
       ? configuredTimeout
-      : REQUEST_TIMEOUT_MS
+      : DEFAULT_REQUEST_TIMEOUT_MS
   const requestRetryCount =
     typeof configuredRetryCount === 'number' && Number.isFinite(configuredRetryCount) && configuredRetryCount >= 0
       ? Math.floor(configuredRetryCount)
-      : REQUEST_RETRY_COUNT
+      : DEFAULT_REQUEST_RETRY_COUNT
   let url: URL
   if (initConfiguration.flaggingProxy?.match('https?://')) {
     // If flaggingProxy has a protocol, use it as-is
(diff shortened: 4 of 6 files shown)


packages/browser/src/transport/fetchConfiguration.ts
```diff
@@ -29,7 +43,77 @@ async function getErrorMessage(response: Response) {
   return response.statusText || 'Unknown error'
 }
 
+function isRetryableStatus(status: number) {
+  return status === 408 || status === 429 || (status >= 500 && status <= 599)
+}
+
+function isRetryableError(error: unknown) {
```
btthomas: Should we make 429s not retryable? i imagine if you hit a 429, waiting 1 second might not help you.

btthomas (approved):

packages/browser/src/domain/configuration.ts
```diff
@@ -71,6 +71,16 @@ export interface FlaggingInitConfiguration extends InitConfiguration {
    * Proxy URL for flagging configuration requests. If set, this will be used instead of the site parameter.
    */
   flaggingProxy?: string
+
+  /**
+   * Timeout for each flag assignment request in milliseconds (default: 1000ms).
+   */
+  assignmentRequestTimeoutMs?: number
```
dd-oleksii: major: I have a vague feeling that 1s is too short for the global default, and the default should either be unlimited or much higher

Our library could be running in all kinds of conditions, like over-populated mobile networks or sparse coverage where you can barely get any response, proxied environments, etc. Having a too short timeout with few retries makes the experience worse

Response times also depend on the complexity and size of the configuration. For configs with thousands of flags, 1s is not unthinkable. If timeout is too short, it could cut off all users

packages/browser/src/transport/fetchConfiguration.ts
```diff
@@ -15,6 +15,20 @@ type JSONAPIError = {
   }[]
 }
 
+const REQUEST_TIMEOUT_MS = 1_000
+const REQUEST_RETRY_COUNT = 1
```
dd-oleksii: nitpick: prefix with `DEFAULT_`

packages/browser/src/transport/fetchConfiguration.ts
```diff
@@ -83,18 +166,25 @@ export function createFlagsConfigurationFetcher(initConfiguration: FlaggingInitC
...
+            response: precomputed,
+            context,
+            fetchedAt: timeStampNow(),
+          },
+        }
+      } catch (error) {
+        if (signal?.aborted || attempt === requestRetryCount || !isRetryableError(error)) {
+          throw error
+        }
+      }
```
dd-oleksii: A more "building blocks"–style approach would be allowing the user to override the fetch function.

So in our code, we call:
```ts
const response = await userFetch(url, params, signal)
const precomputed = await response.json()
```
and all the retries and timeouts are handled inside the fetch function

Our default implementation can have retries/timeout (or maybe not—bundle size), and we can expose helpers/wrappers:
```ts
function fetchWithTimeout(timout, fetch);
function fetchWithRetry(retries, delay, fetch);
```

This would also cover proxying, custom headers, and other things we have special-cased right now.

wdyt?

packages/browser/src/transport/fetchConfiguration.ts
dd-oleksii: 429 is a canonical example of a retryable request: you re-send the request later and get a proper response. It's a good point that 1 second might not help though, so maybe we should respect [`Retry-After`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Retry-After)

packages/browser/src/domain/configuration.ts
sameerank: Also, how much symmetry do we want among the client SDKs?

Server defaults are 30s, which is too long for clients. For mobile, 1s is likely going to result in a lot of timeouts. I currently have this at 5s in https://github.com/DataDog/dd-sdk-ios/pull/3114

Also in the iOS PR, the timeout deadline `initializationTimeout` is for all processes involved in initialization/reconciling the context being set (cache loading, request construction/networking, decoding, and state update). While here, `assignmentRequestTimeoutMs` indicates a bound for the network request, though also includes parsing. I'm also wondering what is worth matching vs allowable to differ here

leoromanovsky: Removed a default value; only customer provided.

packages/browser/src/transport/fetchConfiguration.ts
leoromanovsky: I think a good idea but I wanted to keep the scope here smaller. Can someone working on the blocks project own that? Will adding this parameter now put us in a tough spot with breaking an API later?

sameerank: #351 is the building-blocks work and adds the caller-supplied `fetch` hook discussed here:

https://github.com/DataDog/openfeature-js-client/commit/4b46e33ded7c50c43830cfeed0a135439cfe8f6a

It can follow #348 additively. Once #348 merges, I’ll restack #351 and preserve its timeout/retry options, applying them around the selected fetch implementation when configured. Callers that do not configure those options can own timeout and retry behavior entirely in their supplied `fetch`.

This avoids a breaking change, although it means the #348 configuration fields become an API we continue supporting.

sameerank: Or if it's not too much effort, you could follow the pattern in #351 to do the fetch override here and I'll adjust the building blocks PR stack

packages/browser/src/transport/fetch.ts
```diff
@@ -0,0 +1,77 @@
...
+}
+
+function isRetryableResponse(response: Response): boolean {
+  return response.status === 408 || response.status === 429 || response.status >= 500
+}
+
+/**
+ * Wraps a Fetch-compatible implementation with immediate retries for transient failures.
+ *
+ * Network errors, HTTP 408, HTTP 429, and HTTP 5xx responses are retried. Caller cancellation is never retried.
```
sameerank: One question: the earlier discussion noted that retrying a `429` should respect `Retry-After`, but `withRetry` currently includes `429` among its immediate retries:

https://github.com/DataDog/openfeature-js-client/pull/348#discussion_r3714503845

Since this helper intentionally does not own a delay/backoff policy, should we omit `429` here? Callers that want rate-limit retries can implement that policy through the custom `fetch` hook.

sameerank (approved):
Just one lingering question, but nothing blocking. Lgtm!

aarsilv:
🤖 **From Codex working with Aaron:**

I reviewed the current head (`fea9bec`). The provider integration accomplishes the core goal: custom fetch selection and signal forwarding are wired correctly, omitting `fetch` preserves the existing behavior, and timeout/retry composition works for the provider's current URL-plus-string-body request shape.

I do not think the new public retry helper is quite ready to merge, however:

1. **[P2] `withRetry` cannot retry a `Request` containing a body.** Every attempt passes the same `Request` instance to the underlying fetch. Native Fetch consumes the body on the first attempt, so the next attempt fails because the request has already been used. I reproduced this with a POST `Request`: the wrapper made two calls, but the retry ended with `TypeError: Cannot construct a Request with a Request object that has already been used.` This does not affect the provider's current request path, but it violates the helper's public `typeof fetch` contract. Please create a replayable request per attempt, or explicitly restrict/document non-replayable bodies, and add coverage for a `Request` with a body.

2. **[P3] Cancellation can race with response-body cleanup.** If the caller aborts while `response.body.cancel()` is pending, the loop invokes `fetchImplementation` once more with the already-aborted signal. I reproduced two calls in this situation, despite the documented guarantee that caller cancellation is never retried. Rechecking the signal immediately before every attempt (or after body cancellation) should close this gap.

Validation otherwise looks good: all current GitHub checks are green, and locally I ran all 14 browser suites (181 tests), browser typecheck, browser build, and lint on the changed code successfully.

**Recommendation:** fix the replayability issue before merging and add the cancellation-race test while touching the retry loop. The only recorded approval also predates the substantial redesign, so the final implementation should receive a fresh approval.

packages/browser/src/transport/fetch.ts
```diff
@@ -0,0 +1,77 @@
...
+      requestSignal?.addEventListener('abort', abortFromRequest, { once: true })
+    }
+
+    const timeout = setTimeout(() => {
+      controller.abort(new DOMException(`The request timed out after ${timeoutMs} ms`, 'TimeoutError'))
+    }, timeoutMs)
+
+    try {
+      return await fetchImplementation(input, { ...init, signal: controller.signal })
+    } finally {
```
aarsilv, line 41: 🤖 **From Claude working with Aaron:**

This `finally` ends both guarantees at headers, not at body completion.

`fetch()` resolves at headers. `clearTimeout` disarms the timer. `removeEventListener` detaches the caller signal from `controller`, which is the signal the request uses. `fetchConfiguration.ts:92` parses the body after this, unprotected.

Measured in headless Chromium, against a server that sends headers then stalls the body:

```
plain fetch + caller signal -> json() rejects AbortError
withTimeout(fetch, 60_000)  -> json() hangs
withTimeout(fetch, 300)     -> json() hangs
```

So the timeout does not bound parsing (FFL-2927 AC 3), and cancellation is lost after headers. Here the wrapper is worse than no wrapper. `provider.ts:254` never settles, so `setProviderAndWait()` never resolves and the cache fallback never runs.

The earlier commit on this branch was correct. `git show 11c8704:packages/browser/src/transport/fetchConfiguration.ts` awaits `response.json()` inside the timed `try`.

Fix: hold the timer and the listener until the body is read. Line 19 and `README.md:60` also claim cancellation is preserved.

packages/browser/src/transport/fetch.ts
```diff
@@ -0,0 +1,77 @@
...
+ */
+export function withRetry(fetchImplementation: Fetch, retries: number): Fetch {
+  validateNonNegativeInteger(retries, 'retries')
+
+  return async (input, init) => {
+    const requestSignal = getRequestSignal(input, init)
+
+    for (let attempt = 0; ; attempt += 1) {
+      try {
+        const response = await fetchImplementation(input, init)
```
aarsilv, line 65: 🤖 **From Claude working with Aaron:**

The retry loop has three defects. Codex reported the first two before merge.

1. **Replay.** Line 65 passes the same `input` each attempt. A `Request` with a body is consumed by attempt 1. Measured: `withRetry(fetch, 1)(new Request(url, {method:'POST', body}))` on a 500 throws `TypeError: ...already been used`, so the caller loses the 500. Stream bodies in `init` fail the same way.
2. **Abort race.** Line 66 checks the signal only before `await response.body?.cancel()`. Measured: abort during the cancel gives 2 calls with `aborted=true`. This breaks the "cancellation is never retried" promise and AC 4.
3. **Blind catch.** Line 70 retries every error. A malformed URL, a CORS or CSP denial, and a rejected `cancel()` all retry. All are futile and delay the cache fallback.

The provider path is safe; it passes a string URL and a string body. `index.ts:7` exports these as general Fetch wrappers, so customer code is not.

packages/browser/src/transport/fetch.ts
```diff
@@ -0,0 +1,77 @@
...
+
+function getRequestSignal(input: Parameters<Fetch>[0], init?: Parameters<Fetch>[1]): AbortSignal | undefined {
+  if (init?.signal) {
+    return init.signal
+  }
+  return typeof Request !== 'undefined' && input instanceof Request ? input.signal : undefined
+}
+
+function validateNonNegativeInteger(value: number, name: string): void {
+  if (!Number.isInteger(value) || value < 0) {
```
aarsilv, line 11: 🤖 **From Claude working with Aaron:**

This validator accepts unsafe values.

`timeoutMs: 0` schedules an immediate timer. It does not disable the timeout.

Browser timers convert the delay to a signed 32-bit integer. `3_600_000_000` becomes negative and clamps to zero. Measured in Chromium: the timer fires after 0 ms. Node warns and uses 1 ms. A customer who sets a large value to mean "no timeout" gets 100% `TimeoutError`, provider `ERROR`, and every flag falling back to its in-code default.

`retries` has no cap. `1e21` passes. `2**53 + 2` passes too, and then `attempt += 1` saturates at `2**53`, so `attempt === retries` is never true and the loop at line 63 cannot terminate.

Fix: define `0` explicitly - I suggest `0` means no timeout, and skip the timer for it. Reject delays above 2147483647. Cap `retries`. Add boundary tests for each.

packages/browser/src/transport/fetch.ts
```diff
@@ -0,0 +1,77 @@
...
+      return await fetchImplementation(input, { ...init, signal: controller.signal })
+    } finally {
+      clearTimeout(timeout)
+      requestSignal?.removeEventListener('abort', abortFromRequest)
+    }
+  }
+}
+
+function isRetryableResponse(response: Response): boolean {
+  return response.status === 408 || response.status >= 500
```
aarsilv, line 49: 🤖 **From Claude working with Aaron:**

Retries are immediate, so a 5xx incident gets a synchronized multiplier.

503 is the status a load balancer sheds with, and it usually carries `Retry-After`. This code retries it at once and ignores the header. A 503 returns in a few milliseconds, so every browser retries in the same millisecond bucket as the failure. The README recipe doubles fleet request rate exactly when headroom is lowest.

Excluding 429 is right, but 429 is not the shed signal in this shape.

Fix: honour `Retry-After` on 503, or add backoff with jitter.

packages/browser/src/domain/configuration.ts
```diff
@@ -72,6 +72,12 @@ export interface FlaggingInitConfiguration extends InitConfiguration {
    * Proxy URL for flagging configuration requests. If set, this will be used instead of the site parameter.
    */
   flaggingProxy?: string
+
+  /**
+   * Fetch implementation used for flag configuration requests. It can customize request handling,
+   * including retries, timeouts, proxying, and headers.
+   */
+  fetch?: typeof globalThis.fetch
```
aarsilv, line 80: 🤖 **From Claude working with Aaron:**

This JSDoc scopes the option correctly. Two things around it do not.

1. The READMEs put `fetch` in the main options block with no scope note. Exposure intake (`startExposuresBatch.ts:18`) and flag-evaluation intake (`flagEvaluations.ts:22`) use `createHttpRequest` from `@datadog/browser-core` and ignore this option. A customer who sets `fetch` to route through a corporate proxy sees flag config succeed and exposures drop silently.
2. `flaggingProxy` and `customHeaders` still apply and layer on top. The implicit contract is "your fetch must forward `init.headers` verbatim". A wrapper that replaces `headers` drops `dd-client-token` and gets a 401 with no clue to the cause.

Fix: state the scope in both READMEs, and document the header contract here.

packages/browser/README.md
```diff
@@ -49,9 +49,34 @@ const provider = new DatadogProvider({
...
 })
 ```
 
+### Request Timeouts and Retries
+
+The package provides Fetch-compatible wrappers for adding a timeout and retries. They preserve the provider's
+cancellation signal and can be composed:
+
+```javascript
+import { DatadogProvider, withRetry, withTimeout } from '@datadog/openfeature-browser'
```
aarsilv, line 64: 🤖 **From Claude working with Aaron:**

Three problems in this section.

1. Users of this repository's CDN bundle cannot use these helpers. `webpack.base.js` sets no library output, so the bundle exposes nothing as module exports. Only `defineGlobal` reaches the page, and `index.ts:13` gives it `DD_FLAGGING.Provider`. Webpack then tree-shakes both wrappers out. Expose them on `DD_FLAGGING`, or label this section npm-only.
2. Line 54 uses `customFetch` before this block defines it. A reader who copies the Configuration block gets a `ReferenceError`.
3. Line 60 says the wrappers preserve the provider's cancellation signal. True for `withRetry`, false for `withTimeout` once headers arrive. See `fetch.ts:41`.

The root `README.md` duplicates this text and needs the same three corrections.

packages/browser/test/transport/fetch.spec.ts
```diff
@@ -0,0 +1,139 @@
...
+      .mockRejectedValueOnce(new TypeError('Network error'))
+      .mockResolvedValueOnce(response)
+
+    await expect(withRetry(fetchImplementation, 1)('/flags')).resolves.toBe(response)
+    expect(fetchImplementation).toHaveBeenCalledTimes(2)
+  })
+
+  it.each([408, 500, 599])('retries an HTTP %s response', async (status) => {
+    const cancel = jest.fn().mockResolvedValue(undefined)
+    const retryableResponse = { status, body: { cancel } } as unknown as Response
```
aarsilv, line 79: 🤖 **From Claude working with Aaron:**

This fake is the only thing that reaches `response.body?.cancel()`. Real streams are never tested.

jsdom supplies no `Response`, so `test-setup.ts:2` installs `whatwg-fetch`, which has no `body` getter. I probed this suite: `'body' in new Response(...)` is `false`. So source line 69 short-circuits for every constructed `Response`.

The fake does guard that the call happens - deleting line 69 fails this test. It does not guard ordering. I moved the `cancel()` above the return check, the natural "always drain" refactor, and all 14 suites and 181 tests still passed. That refactor consumes the final 5xx body, so `getErrorMessage` (`fetchConfiguration.ts:89`) would throw a stream error instead of returning the server's detail.

Fix, in two parts. First add `/** @jest-environment node */` to this file. Node supplies `fetch`, `Response`, `ReadableStream`, `DOMException`, `AbortController`, and `Request`, and `whatwg-fetch` then leaves those globals alone. All 17 existing cases pass unchanged; I verified this.

Then replace the fake with a native `Response`:

```ts
it('cancels the retryable response body before the next attempt', async () => {
  let cancelled: unknown = 'not cancelled'
  const stream = new ReadableStream({
    start: (controller) => controller.enqueue(new TextEncoder().encode('{"errors":[]}')),
    cancel: (reason) => { cancelled = reason ?? 'cancelled' },
  })
  const retryable = new Response(stream, { status: 503 })
  const successful = new Response(null, { status: 200 })
  const fetchImplementation = jest.fn().mockResolvedValueOnce(retryable).mockResolvedValueOnce(successful)

  await expect(withRetry(fetchImplementation, 1)('/flags')).resolves.toBe(successful)

  expect(cancelled).toBe('cancelled')
  expect(retryable.bodyUsed).toBe(true)
})
```

And add the case that catches the ordering regression - assert the final 5xx keeps `bodyUsed === false` and stays parseable after retries are exhausted. Both tests pass as merged and fail on their respective regressions.

CHANGELOG.md
```diff
@@ -6,6 +6,12 @@
 
 ---
 
+## Unreleased
```
aarsilv, line 9: 🤖 **From Claude working with Aaron:**

This heading breaks the file's convention. Every other section is `## @datadog/<pkg>@<version>`, and every bullet back to #275 carries a `([#NNN](...))` link.

Two effects: the change is not traceable to a PR, and `## Unreleased` is now the first `##`, so the generator splices above it and leaves it orphaned.

Fix: retitle to `## @datadog/openfeature-browser@1.3.0` and add the #348 link.

PR #349, opened 2026-08-04
[FFL-2926] Add tracking parity to DatadogOfflineProvider

sameerank:
## Summary

- add optional browser telemetry configuration to `DatadogOfflineProvider`
- reuse the existing `DatadogProvider` exposure, flag-evaluation, and RUM tracking hooks through shared setup
- keep offline evaluation telemetry-free when `tracking` is omitted
- keep configuration delivery through `setConfiguration()`, consistent with React Native
- attribute telemetry to the actual effective OpenFeature context used for evaluation
- support the same tracking behavior for precomputed and rules-based offline evaluation
- reset exposure deduplication when `setConfiguration` replaces the active configuration
- prevent asynchronous persistent-cache initialization from restoring stale deduplication entries after replacement
- keep offline configuration delivery free of fetching and polling even when telemetry is enabled

## Why

Offline describes configuration delivery, not necessarily telemetry. Consumers should be able to opt into the same browser tracking behavior as `DatadogProvider` without introducing configuration fetching or polling.

When tracking is omitted, `DatadogOfflineProvider` remains evaluation-only and creates no tracking hooks or telemetry requests. The constructor accepts only optional browser tracking settings; caller-supplied flag configuration is loaded separately with `setConfiguration()`.

Jira: [FFL-2926](https://datadoghq.atlassian.net/browse/FFL-2926)

## Behavior and compatibility

- supplying a `tracking` block enables exposure, flag-evaluation, and RUM tracking with the same defaults and hook order as `DatadogProvider`
- each integration can be disabled independently
- context-specific precomputed consumers must register the matching context; `{}` is treated literally
- tracking hooks observe the same OpenFeature context used for evaluation
- successful precomputed and rules evaluations carry existing allocation, variation, context, logging, and timestamp metadata
- defaults and failed evaluations do not emit exposure events
- tracking failures do not affect local evaluation or provider lifecycle
- the existing `DatadogProvider` constructor, fetching, caching, and tracking behavior remain unchanged

## Stack and merge order

This PR is stacked on [#346](https://github.com/DataDog/openfeature-js-client/pull/346), which documents static initialization with `getPrecomputedContext` and treats empty OpenFeature contexts literally.

1. Merge [#344](https://github.com/DataDog/openfeature-js-client/pull/344).
2. Merge [#353](https://github.com/DataDog/openfeature-js-client/pull/353).
3. Retarget and merge [#336](https://github.com/DataDog/openfeature-js-client/pull/336).
4. Retarget and merge [#346](https://github.com/DataDog/openfeature-js-client/pull/346).
5. Retarget this PR to `main` and merge it.

## Validation

- full browser suite passed: 18 suites and 203 tests
- Biome checks passed

[FFL-2926]: https://datadoghq.atlassian.net/browse/FFL-2926


sameerank:
If a building blocks design seems like an overly complicated dev experience for tracking, then I can revert back to https://github.com/DataDog/openfeature-js-client/pull/349/commits/ff35a7d6ecf7710bb7691816fc637e900ef5ea2a where we abstract away the exposure cache initialization and reset implementation details, e.g.

```javascript
const provider = new DatadogOfflineProvider({
  tracking: {
    clientToken: 'client-token',
    applicationId: 'application-id',
    site: 'datadoghq.com',
    service: 'storefront',
    enableRumFeatureFlagTracking: false,
  },
})
provider.setConfiguration(configuration)
```

sameerank:
One cache-management nuance worth calling out explicitly:

`DatadogCoreProvider` + manually registered exposure hooks intentionally does not manage the exposure cache exactly the same way as `DatadogProvider`.

`DatadogProvider` owns both configuration delivery and its provider-managed exposure hook/cache, so during provider lifecycle changes it can clear the exposure cache when the fetched precomputed configuration changes.

`DatadogCoreProvider` is evaluation-only and does not own registered hooks, so it does not clear hook state directly. Instead, when `setConfiguration()` installs a configuration, the provider computes an internal configuration identity and stamps exposure-eligible evaluation metadata with it. The exposure hook includes that identity in the cached assignment value. That gives the behavior we need without adding provider-owned side effects:

- same config identity: repeated exposure remains deduped
- changed config identity: the same flag/user/allocation can emit again
- cache storage/backing behavior still matches the existing browser exposure cache path: Chrome storage when available, then `localStorage`, then memory

There is still no explicit SDK-level max size or TTL on the browser exposure cache, which matches the existing `DatadogProvider` cache-size behavior. The difference is the reset mechanism: `DatadogProvider` can clear its owned cache during lifecycle changes, while `DatadogCoreProvider` uses config-scoped dedup values because hooks are registered manually and remain application-managed.

packages/browser/src/openfeature/core-provider-metadata.ts
```diff
@@ -0,0 +1,31 @@
+import type { EvaluationDetails, FlagValue, ResolutionDetails } from '@openfeature/web-sdk'
+
+const CORE_CONFIGURATION_ID_METADATA_KEY = '__dd_core_configuration_id'
+
+export function withCoreConfigurationId<T extends FlagValue>(
```
sameerank: Cache-management note: this is intentionally different from `DatadogProvider`'s provider-owned exposure cache reset. `DatadogCoreProvider` does not own manually registered hooks, so it should not clear hook state during provider lifecycle changes. Instead, it stamps exposure-eligible evaluations with the active configuration identity, and the exposure hook includes that identity in the cached assignment value. That keeps repeated exposures deduped for the same configuration while allowing the same flag/user/allocation to emit again after `setConfiguration()` installs a changed configuration. The backing cache behavior is still the same browser path as `DatadogProvider` (Chrome storage, then `localStorage`, then memory) and still has no explicit SDK-level max size or TTL.

packages/browser/src/openfeature/core-provider.ts
```diff
@@ -1,69 +1,100 @@
...
+    )
+  }
+
+  private canEvaluateCurrentContext(): boolean {
+    return this.context !== undefined && !getFlagsConfigurationError(this.flagsConfiguration, this.context)
+  }
+
+  private computeConfigurationId(configuration: FlagsConfiguration): string {
+    try {
+      return getMD5Hash(configurationToString(configuration))
```
btthomas: Did a review w/ Astra:

> Unchanged fetched configurations resend exposures
> DatadogCoreProvider.setConfiguration() hashes the complete serialized configuration as its identity. That serialization includes fetchedAt, and both public fetch helpers generate a new fetchedAt on every request.
> This would affect any client that would have a refetch rules mechanism (on old sessions, after some time where the rules might be stale)
  1. The application fetches for rules a second (or nth time)
  2. The response is unchanged.
  3. Only fetchedAt changes.
  4. The provider generates a different configuration identity.
  5. The same assignment emits another exposure.
> Recommended fix: derive identity from stable configuration content or a trusted content revision, excluding retrieval timestamps. Add rules and precomputed tests where only fetchedAt changes.

Do we have some incrementing "version" of a UFC available? That would be constant if the rules do not change and would be exactly what we want, right?

packages/browser/src/openfeature/tracking.ts
```diff
@@ -0,0 +1,42 @@
...
+
+type TrackingToggleOptions =
+  | 'enableExposureLogging'
+  | 'enableFlagEvaluationTracking'
+  | 'enableRumFeatureFlagTracking'
+  | 'rum'
+
+export type DatadogTrackingHooksOptions = Omit<FlaggingTrackingInitConfiguration, TrackingToggleOptions>
+
+export interface DatadogTrackingHooks {
```
btthomas: > Tracking hooks cannot be disposed
>   Creating an evaluation hook immediately:
  - Starts a recurring aggregation interval.
  - Creates batch infrastructure.
  - Subscribes to page-exit events.
>  However, DatadogTrackingHook and DatadogTrackingHooks expose only initialize(). Removing the hooks or clearing the provider does not stop those resources.
> after client.clearHooks() and OpenFeature.clearProviders(), the removed hook still sends its queued evaluation batch and retained its recurring timer.

>Recommended fix: add an idempotent shutdown()/dispose() lifecycle, propagate it through createDatadogTrackingHooks(), and stop aggregators, batches, and subscriptions. Prefer starting resources during initialize(), not factory construction.

Providers have "shutdown" and we can propagate that to the hooks, so they stop listening: https://openfeature.dev/specification/sections/providers#25-shutdown

packages/browser/src/openfeature/exposures.ts
```diff
@@ -59,3 +78,36 @@ export function createExposureLoggingHook(
...
+export function createDatadogExposureLoggingHook(options: DatadogTrackingHooksOptions): DatadogExposureLoggingHook {
+  const configuration = validateAndBuildFlaggingTrackingConfiguration(options)
+  if (!configuration) {
+    return {
+      hooks: [],
+      initialize: () => Promise.resolve(),
+    }
+  }
+
+  const exposureCache = assignmentCacheFactory({
```
btthomas: > Different tracking destinations share one exposure cache

>   Every standalone exposure hook uses the same persistent cache namespace: dd-of-browser. The assignment-cache key contains only the flag and subject; neither the destination, client token, environment, nor application is included.
  1. Hook A uses token/application A and records an exposure.
  2. Hook B is created later with token/application B.
  3. Hook B evaluates the same configuration, flag, and subject.
  4. The flag evaluates correctly, but B sends no exposure because A’s persistent record suppresses it.

> This also affects changing telemetry destinations after deployment. The online provider already has a related shared-cache weakness, but exposing independently composable factories makes the behavior part of the new public API.

> Recommended fix: namespace persistent caches by a non-secret hash of the effective telemetry scope, including destination/account, environment, and application. Apply this to both localStorage and Chrome storage. Add multi-destination and reinitialization tests.

This would affect a customer with extra custom logging (maybe sending exposures to their own warehouse). We should also double check to make sure this doesn't affect the batteries included Datadog Provider.

btthomas (commented):
Overall it looks close.

Found three things from AI.

leoromanovsky (approved):
Minor improvement if you'd like to accept now or later: https://github.com/DataDog/openfeature-js-client/pull/391

packages/browser/src/openfeature/tracking.ts
sameerank: Addressed in https://github.com/DataDog/openfeature-js-client/commit/3d01ce3475c75d4e59eb92fe1994c2c10cdc4ef2. Exposure/evaluation resources now start during `initialize()`, and both their individual controllers and the combined controller expose an idempotent `shutdown()`. Shutdown flushes pending events, stops timers, and removes subscriptions; already-submitted transport requests may still complete or retry.

The regular `DatadogProvider` propagates its OpenFeature `onClose()` lifecycle to its built-in tracking. For `DatadogCoreProvider`, manually registered client hooks remain application-owned, so applications unregister them and call `tracking.shutdown()` explicitly. Clearing hooks or removing the core provider alone does not dispose of them.

Added tests for lazy startup, repeated shutdown/reinitialization, shutdown during pending initialization, and provider-owned cleanup.

packages/browser/src/openfeature/core-provider.ts
sameerank: I checked the UFC backend implementation and didn’t find a configuration-wide incrementing revision. There is a semantic `Fingerprint()` used for ETags, which excludes the per-build `CreatedAt` timestamp.

The browser fetch helper doesn’t currently retain that ETag, and application-supplied configurations may not have one, so I kept the locally computed content identity. In https://github.com/DataDog/openfeature-js-client/commit/562cf24049d9d13522f70accf45a98b74f24da10, it excludes `fetchedAt`/`etag` and, for rules configurations, `createdAt`. These fields remain unchanged in the configuration and portable wire representation.

Added regression coverage for rules and precomputed refetches, timestamp-only rules rebuilds, and actual content changes that should allow a new exposure.

packages/browser/src/openfeature/exposures.ts
sameerank: Confirmed this also affected the regular `DatadogProvider`. https://github.com/DataDog/openfeature-js-client/commit/885716b397bc93d6ddc3a966c679aad29483c95f applies the same telemetry-scoped cache construction to both it and the standalone exposure hook.

The persistent namespace hashes the effective telemetry scope, including destination, client token, environment, application, service, and source. Both localStorage and Chrome storage are isolated, and clearing Chrome storage removes only the relevant namespace. Function-valued proxies use memory-only deduplication because their destination cannot reliably be inferred.

I also replaced the inherited `eppo-assignment` prefix with `datadog-assignment` in both storage implementations. Added multi-scope and recreation tests for both provider paths and storage backends, plus assertions for the persisted namespace names.

**Upgrading starts a fresh cache. Previously cached assignments may emit another exposure on their next evaluation**; subsequent identical evaluations are deduplicated. We cannot safely migrate legacy entries because they do not identify the telemetry destination, so reusing them could preserve the cross-destination suppression bug.

For Datadog experiment analysis, the [assignment query](https://github.com/DataDog/dd-source/blob/2080e97df96802d7f6dbcde42338aa3c5a711b37/domains/ffe/libs/experimentation/pipeline/common/common_blocks.go#L785-L820) collapses exposures by subject and retains the first exposure. Repeated exposures for the same subject/variant within the analysis window therefore should not add participants.

PR #351, opened 2026-08-04
[FFL-2934] Add browser configuration fetchers

sameerank:
## Summary

- add public browser fetchers for precomputed and rules-based flag configurations
- keep browser-specific HTTP and authentication concerns in the browser package and protobuf decoding in flagging core
- send `dd-application-id` only to the precomputed endpoint; rules-based requests use `dd-client-token` and the protobuf `Accept` header
- support ETag conditional requests and reuse prior configurations on `304 Not Modified`
- keep the `/precomputed` entry point free of protobuf parsing dependencies
- preserve the existing `DatadogProvider.initialFlagsConfiguration` public API as a network-failure fallback
- use `DatadogOfflineProvider.setConfiguration()` for explicit caller-managed configuration

## User impact

Consumers can fetch portable configuration explicitly and load it into `DatadogOfflineProvider`:

```ts
const configuration = await fetchPrecomputedConfiguration(options)
const provider = new DatadogOfflineProvider()
provider.setConfiguration(configuration)
```

The same offline-provider flow works with `fetchRulesConfiguration`.

`DatadogProvider` remains network-first and continues to fetch precomputed configuration during initialization and context changes. For backward compatibility, it still accepts `initialFlagsConfiguration`; a matching precomputed configuration remains available as the in-memory fallback if the network request fails. Applications that want explicit offline configuration ownership can use `DatadogOfflineProvider` and transition to the online provider through provider composition or handoff once it is ready.

This matches the React Native separation between offline configuration delivery and online fetching while leaving transition policy outside either provider.

## Stack and merge order

This PR is stacked on [#346](https://github.com/DataDog/openfeature-js-client/pull/346). Its documented precomputed initialization flow uses `getPrecomputedContext` from [#353](https://github.com/DataDog/openfeature-js-client/pull/353).

1. Merge [#344](https://github.com/DataDog/openfeature-js-client/pull/344).
2. Merge [#353](https://github.com/DataDog/openfeature-js-client/pull/353).
3. Retarget and merge [#336](https://github.com/DataDog/openfeature-js-client/pull/336).
4. Retarget and merge [#346](https://github.com/DataDog/openfeature-js-client/pull/346).
5. Retarget this PR and sibling [#349](https://github.com/DataDog/openfeature-js-client/pull/349) to `main`; they can merge independently.

[Jira: FFL-2934](https://datadoghq.atlassian.net/browse/FFL-2934)

## Validation

- full browser suite passed: 16 suites and 195 tests
- browser TypeScript checks, Biome checks, and production build passed
- regression coverage verifies `initialFlagsConfiguration` remains an in-memory fallback after a network failure
- direct precomputed and rules-based fetching were verified in the browser
- a live rules-based protobuf response decoded successfully and all returned flags were exercised through core evaluation and `DatadogOfflineProvider` without evaluation errors


packages/browser/src/transport/fetchConfiguration.ts
```diff
@@ -29,72 +47,118 @@ async function getErrorMessage(response: Response) {
...
+  const defaultHeaders = buildConfigurationHeaders(
+    options,
+    {
+      'Content-Type': 'application/vnd.api+json',
+      ...(options.previousConfiguration?.precomputed?.etag && {
+        'If-None-Match': options.previousConfiguration.precomputed.etag,
+      }),
+    },
+    'precomputed'
+  )
```
leoromanovsky, line 112: Hey I was looking at this function for correctness. I didn't trust the codex review without writing my own test to prove it, and at that point I decided I might as well add a fix to. PTAL and merge if accepted: https://github.com/DataDog/openfeature-js-client/pull/364

leoromanovsky (approved):
Thought about etag some more and since the backend function doesn't return one anyways, nor has support for preflight we should punt on it now anyways. Do you want this support added? In terms of billing it'll still be a request to the CDN so we aren't saving customers any money but if the preflight succeeds their application will startup faster with the full network wait.

packages/browser/src/transport/fetchConfiguration.ts
```diff
@@ -80,6 +80,10 @@ export function buildConfigurationHeaders(
       : {
           'dd-client-token': options.clientToken,
           ...(endpoint === 'precomputed' && options.applicationId && { 'dd-application-id': options.applicationId }),
+          ...(endpoint === 'rules' && {
+            'DD-Client-Library-Language': sourcePayload.sdk_name,
+            'DD-Client-Library-Version': sourcePayload.sdk_version,
```
sameerank, line 94: Used as cache keys according to https://github.com/DataDog/terraform-config/blob/42167906c0861641601b9ba73c8b8d96d9609417/domains/ffe/modules/ufc-vcl-service/main.tf#L181-L186

And allowed in the header according to https://github.com/DataDog/terraform-config/blob/51751e4541a2e1814bf804a90465d4e39b5ad197/domains/ffe/modules/ufc-vcl-service/main.tf#L166

packages/browser/src/transport/fetchConfiguration.ts
```diff
@@ -29,72 +47,122 @@ async function getErrorMessage(response: Response) {
...
-    ...initConfiguration.customHeaders,
+    ...options.customHeaders,
   }
+}
+
+export async function fetchPrecomputedConfiguration(
+  options: PrecomputedConfigurationFetchOptions
+): Promise<FlagsConfiguration> {
+  const url = buildConfigurationUrl(options, 'precomputed')
+  const defaultHeaders = buildConfigurationHeaders(
```
btthomas, line 106: >   Both fetchers send If-None-Match, but the live CORS preflight responses do not permit that header:
>  - Precomputed permits Content-Type, dd-client-token, dd-application-id, and X-rkyv.
>  - Rules permits the token, Accept, language, and version headers.
>  - Neither successful response path exposes ETag to browser JavaScript.
>  - The precomputed edge implementation does not currently return or process an ETag.

Do we need to make the change to edge assignments service?
https://github.com/ddoghq/dd-source/blob/1a3abd5e4dc749e15bb9c3ac82c488ed522ac32c/domains/ffe/apps/edge-assignments/src/main.rs#L81-L88

packages/browser/src/transport/fetchConfiguration.ts
```diff
@@ -29,72 +47,122 @@ async function getErrorMessage(response: Response) {
...
-        response: precomputed,
-        context,
-        fetchedAt: timeStampNow(),
       },
-    }
+    }),
+  })
+  if (response.status === 304 && options.previousConfiguration?.precomputed) {
+    return { precomputed: options.previousConfiguration.precomputed }
+  }
```
btthomas: > The 304 handler must not return the previous configuration unchanged. The previous configuration contains the old context. Doing so causes configMatchesContext() to reject the assignments later.
> Instead, it should reuse the response but bind it to the requested context:

Suggested fix:
```
  if (response.status === 304 && previousConfiguration?.precomputed) {
    return {
      precomputed: {
        ...previousConfiguration.precomputed,
        context: options.context,
        fetchedAt: timeStampNow(),
      },
    }
  }
```

packages/browser/src/transport/fetchConfiguration.ts
```diff
@@ -29,72 +47,122 @@ async function getErrorMessage(response: Response) {
...
+  if (!response.ok) {
+    const errorMessage = await getErrorMessage(response)
+    throw new Error(`Failed to fetch flag configuration: ${errorMessage}`)
+  }
+  const precomputed = await response.json()
```
btthomas: > A successful HTTP response is converted directly into FlagsConfiguration. The PR explicitly recommends passing that value into DatadogOfflineProvider, but neither the fetcher nor setConfiguration() validates the response structure. If we get back `{}` The offline provider will throw for `undefined (reading 'attributes')`.

Maybe we need a validation here and some tests for `null`, `{}`, etc.

btthomas (commented):
Some AI reviews and my comments. Getting close!

packages/browser/src/transport/fetchConfiguration.ts
sameerank: Thanks for putting this together. Oleksii clarified that ETags should represent the generated response. Reusing an ETag across contexts is safe when both contexts produce the same response; if the assignments differ, the server must return a different ETag.

Since the live endpoints don’t currently expose ETags or support `If-None-Match` through CORS, I’m going to remove ETag/304 handling from #351 and defer this. When we add it back, a `304` should reuse the response while binding it to the newly requested context.

sameerank: https://github.com/DataDog/openfeature-js-client/pull/351/commits/088d46506788b6b46cf3b6b3fc5e4cd2d746db5a

packages/browser/src/transport/fetchConfiguration.ts
sameerank: Since the live endpoints don’t currently expose ETags or permit `If-None-Match` through CORS, I removed ETag/`304` handling from both fetchers rather than expanding the edge work in this PR:

https://github.com/DataDog/openfeature-js-client/commit/088d46506788b6b46cf3b6b3fc5e4cd2d746db5a

We can reintroduce it later as a coordinated backend and SDK change.

packages/browser/src/transport/fetchConfiguration.ts
sameerank: The `304` path has been removed along with ETag support, so the fetcher can no longer reuse a configuration carrying the previous context:

https://github.com/DataDog/openfeature-js-client/commit/088d46506788b6b46cf3b6b3fc5e4cd2d746db5a

If conditional requests are added later, a `304` should reuse the response while binding it to the requested context.

packages/browser/src/transport/fetchConfiguration.ts
sameerank: Makes sense. An HTTP success does not guarantee that the response is a valid precomputed configuration.

Addressed in https://github.com/DataDog/openfeature-js-client/commit/dfc299c9baafc65bc0cfa62dd5841061ec5f97c6

`fetchPrecomputedConfiguration` now validates the response through the shared core parser. Configuration-level failures such as `null` and `{}` return `precomputedError`, so evaluation reports `PARSE_ERROR` instead of throwing. Malformed individual flags are recorded in `flagErrors` while valid sibling flags remain usable. I added regression coverage for all three cases.

btthomas (approved):

packages/browser/src/openfeature/provider.ts
```diff
@@ -167,7 +167,9 @@ export class DatadogProvider extends DatadogCoreProvider {
     // Important: OF SDK awaits for all onContextChange calls to exit
     // before marking the provider as ready. Make sure to respect
     // `signal`, so we don't block OF SDK unnecessarily.
-    this.latestContextUpdate = this.retrieveFlagsConfiguration(evaluationContext, { signal })
+    const configurationPromise = this.retrieveFlagsConfiguration(evaluationContext, { signal })
+
+    this.latestContextUpdate = configurationPromise
```
dd-oleksii: nitpick: seems to be an unrelated change?

packages/browser/src/transport/fetchConfiguration.ts
```diff
@@ -15,10 +15,27 @@ type JSONAPIError = {
   }[]
 }
 
-async function getErrorMessage(response: Response) {
+export interface ConfigurationFetchOptions {
+  clientToken: string
+  applicationId?: string
+  env?: string | null
```
dd-oleksii, line 21: major: I believe `env` should be required?

packages/browser/src/transport/fetchConfiguration.ts
```diff
@@ -15,10 +15,27 @@ type JSONAPIError = {
...
+  site?: string
+  customHeaders?: Record<string, string>
+  overwriteRequestHeaders?: boolean
+  flaggingProxy?: string
```
dd-oleksii, line 25: major (given this is now part of our public API): we've chatted with Leo in a PR adding polling and retries that it could be simpler and more flexible to allow users to pass their own `fetch` function instead of providing tons of configuration options. This makes the interface and our implementation simpler, as well as allow further customization (e.g. overriding URL, adding retries etc)

packages/browser/src/index.ts
```diff
@@ -1,5 +1,8 @@
 import { registerGlobal } from './register-global'
 
+export type { ConfigurationFetchOptions, PrecomputedConfigurationFetchOptions } from './transport/fetchConfiguration'
```
dd-oleksii: major: I believe we only need to exported `PrecomputedConfigurationFetchOptions`. There's no place our users can plug `ConfigurationFetchOptions` into

packages/browser/src/transport/fetchConfiguration.ts
```diff
@@ -29,72 +46,112 @@ async function getErrorMessage(response: Response) {
...
+    ...(options.overwriteRequestHeaders
       ? {}
       : {
-          'dd-client-token': initConfiguration.clientToken,
-          ...(initConfiguration.applicationId && { 'dd-application-id': initConfiguration.applicationId }),
+          'dd-client-token': options.clientToken,
+          ...(endpoint === 'precomputed' && options.applicationId && { 'dd-application-id': options.applicationId }),
+          ...(endpoint === 'rules' && {
+            'DD-Client-Library-Language': sourcePayload.sdk_name,
+            'DD-Client-Library-Version': sourcePayload.sdk_version,
```
dd-oleksii, line 94: minor: shall we send these to precomputed endpoint as well? they are helpful

EDIT: ok, it looks like we'd need to modify CORS first, so feel free to ignore

packages/browser/src/transport/fetchRulesConfiguration.ts
```diff
@@ -0,0 +1,34 @@
...
+    signal: options.signal,
+  })
+  if (!response.ok) {
+    const errorMessage = await getErrorMessage(response)
+    throw new Error(`Failed to fetch flag configuration: ${errorMessage}`)
+  }
+
+  const configuration = configurationFromRulesBinary(new Uint8Array(await response.arrayBuffer()))
+  if (configuration.rules) {
+    configuration.rules.fetchedAt = timeStampNow()
```
dd-oleksii: nitpick: I think the `fetchedAt` timestamp should be stored before we make the request. It's primarily going to be used to determine whether the client needs to refresh their configuration or the current one is fresh enough. Using an earlier timestamp is more conservative

packages/browser/src/rules-based.ts
```diff
@@ -9,3 +9,10 @@ export {
   getPrecomputedContext,
 } from '@datadog/flagging-core/rules-based'
 export * from './provider-entrypoint'
+export type {
+  ConfigurationFetchOptions,
```
dd-oleksii: major: similarly here, I don't think we need to export `ConfigurationFetchOptions`

packages/core/src/configuration/index.ts
```diff
@@ -10,3 +10,5 @@ export {
   configurationToPrecomputedString as configurationToString,
 } from './precomputed-wire'
 export type { FlagsConfigurationWire } from './wire-types'
+/** @internal */
+export { parsePrecomputedConfigurationResponse } from './wire-validation'
```
dd-oleksii: minor: hm... does this attach `@internal` to `parsePrecomputedConfigurationResponse` correctly? shall we have this comment on the function definition?

packages/core/src/configuration/ufc-protobuf.ts
```diff
@@ -5,7 +5,11 @@ import { base64Decode, base64Encode } from '@bufbuild/protobuf/wire'
 import { type FlagsConfiguration, FlagsConfigurationSchema } from './generated/ufc_pb'
 
 export function decodeUniversalFlagConfiguration(response: string): FlagsConfiguration {
-  return fromBinary(FlagsConfigurationSchema, base64Decode(response))
+  return decodeUniversalFlagConfigurationBinary(base64Decode(response))
+}
+
+export function decodeUniversalFlagConfigurationBinary(response: Uint8Array): FlagsConfiguration {
```
dd-oleksii: nitpick: very opinionated but I'm trying to gradually phase out "universal flag configuration" usage in favor of simple "flags configuration." We have so many different configuration formats that calling them all "universal" doesn't make sense

packages/core/src/configuration/ufc-protobuf.ts
```diff
@@ -5,7 +5,11 @@ import { base64Decode, base64Encode } from '@bufbuild/protobuf/wire'
...
+  return fromBinary(FlagsConfigurationSchema, response)
 }
```
dd-oleksii: minor: imo "protobuf" module should not care about base64 encoding/decoding the whole response. Protobuf _is_ binary-only.

Base64 is part of configuration wire spec, so base64 encoding/decoding should happen there

packages/browser/src/openfeature/provider.ts
sameerank: Yup unrelated and unnecessary. Removed in https://github.com/DataDog/openfeature-js-client/commit/c4f57320f04a2c91f85a5fda068d84ccbf0568cd

packages/browser/src/transport/fetchConfiguration.ts
sameerank: Made `env` required for both standalone fetch option types in https://github.com/DataDog/openfeature-js-client/commit/4b46e33ded7c50c43830cfeed0a135439cfe8f6a

We can do this because these new standalone APIs have no existing compatibility constraint. However, `FlaggingInitConfiguration.env` remains optional because that is an existing public provider API.

https://github.com/DataDog/openfeature-js-client/blob/135656656c8e7b873059b2e346ba0070e8f13bdf/packages/browser/src/domain/configuration.ts#L11

packages/browser/src/transport/fetchConfiguration.ts
sameerank: Implemented in https://github.com/DataDog/openfeature-js-client/commit/4b46e33ded7c50c43830cfeed0a135439cfe8f6a

Both standalone fetchers and `DatadogProvider` can now use a caller-supplied `fetch` implementation, allowing it to own retries, timeouts, proxying, or header customization.

The new standalone option types do not expose the proxy/header-specific fields. I retained them only on `FlaggingInitConfiguration` because they were already part of the existing provider API:

https://github.com/DataDog/openfeature-js-client/blob/02461e383bc0433200caa392eec6b4f9b3c8abbe/packages/browser/src/domain/configuration.ts#L49-L62

packages/browser/src/index.ts
sameerank: Removed the `ConfigurationFetchOptions` export in https://github.com/DataDog/openfeature-js-client/commit/4b46e33ded7c50c43830cfeed0a135439cfe8f6a

The default entry point now exports only `PrecomputedConfigurationFetchOptions`.

packages/browser/src/transport/fetchConfiguration.ts
sameerank: Agreed, and I’ll leave these headers rules-only for now (until the precomputed endpoint’s CORS configuration allows them)

packages/browser/src/transport/fetchRulesConfiguration.ts
sameerank: Done in https://github.com/DataDog/openfeature-js-client/commit/973dd015a14b999c1b26627818a1e22611844541

Both fetchers now capture `fetchedAt` before starting the request. I also added regression assertions verifying that the timestamp has already been recorded when the supplied fetch implementation is invoked.

packages/browser/src/rules-based.ts
sameerank: Removed the base type export in https://github.com/DataDog/openfeature-js-client/commit/4b46e33ded7c50c43830cfeed0a135439cfe8f6a

The rules-based entry point now exports only the capability-specific `PrecomputedConfigurationFetchOptions` and `RulesConfigurationFetchOptions`.

packages/core/src/configuration/index.ts
sameerank: Good catch. `@internal` is now attached directly to the function definition:

https://github.com/DataDog/openfeature-js-client/blob/ddcd83076a85327c5e2f6794cebf1eac39bd5492/packages/core/src/configuration/wire-validation.ts#L27-L28

I also removed the redundant annotation from the re-export:

https://github.com/DataDog/openfeature-js-client/commit/2c9e9fceba7516066287684ef728efa221f98570

packages/core/src/configuration/ufc-protobuf.ts
sameerank: Sounds good! Renamed the protobuf helpers from `decode/encodeUniversalFlagConfiguration` to `decode/encodeFlagsConfiguration` and updated the related terminology.

https://github.com/DataDog/openfeature-js-client/commit/d7797c5ed35914ff566601b78bd3935f153e1957

packages/core/src/configuration/ufc-protobuf.ts
sameerank: The protobuf codec now accepts and returns binary data only. Base64 decoding/encoding happens at the configuration wire boundary, while binary endpoint responses pass directly to the protobuf decoder.

https://github.com/DataDog/openfeature-js-client/commit/73c493caed181ba3dfba8ba772ac2abac5cd537b

PR #353, opened 2026-08-05
[FFL-2837] Expose precomputed configuration context

btthomas:
## Stack

1. [Parse flags configuration wire format: #344](https://github.com/DataDog/openfeature-js-client/pull/344)
2. Expose the precomputed context <- you are here

This PR uses #344 as its base. Merge #344 first. Then retarget this PR to `main`.

## Summary

This PR adds `getPrecomputedContext(configuration)` to `@datadog/flagging-core`.

The function returns a detached copy of the context from a valid precomputed configuration. It returns `undefined` when the configuration does not have a context-specific precomputed branch. The function does not change the configuration, OpenFeature, or provider state.

The function is available from these entry points:

- `@datadog/flagging-core`
- `@datadog/flagging-core/configuration`
- `@datadog/flagging-core/precomputed`
- `@datadog/openfeature-browser`
- `@datadog/openfeature-browser/precomputed`

The precomputed entry points remain free of Protobuf-ES modules.

## Reason

The configuration wire and parsed configuration are opaque contracts. SDK users need a supported way to set the OpenFeature context for a context-specific precomputed configuration.

Without this function, each SDK must inspect internal configuration fields or implement the same accessor. A shared function keeps browser and React Native behavior consistent.

## Behavior

- Preserve an explicit empty context.
- Preserve an explicit empty targeting key.
- Deep-copy nested objects, arrays, and dates.
- Return `undefined` for empty, rules-only, invalid-precomputed, and context-agnostic configurations.
- Return the precomputed context from a mixed valid configuration.

## Validation

- 435 flagging-core tests passed.
- 156 browser tests passed.
- Flagging-core and browser type checks passed.
- Flagging-core and browser builds passed.
- Biome and Prettier checks passed.


packages/browser/README.md
```diff
@@ -95,17 +95,34 @@ console.log(result.reason) // Evaluation reason
...
+  await OpenFeature.setContext(context)
+}
+
+const provider = new DatadogProvider({
+  clientToken: 'pub_...',
+  site: 'datadoghq.com',
+  env: 'production',
+  initialFlagsConfiguration: configuration,
+})
+await OpenFeature.setProvider(provider)
```
sameerank: `.setProvider` is not awaitable so

```suggestion
await OpenFeature.setProviderAndWait(provider)
```

or if we want to show an atomic registration of the context and initialization

```suggestion
if (context !== undefined) {
  await OpenFeature.setProviderAndWait(provider, context)
} else {
  await OpenFeature.setProviderAndWait(provider)
}
```

packages/browser/README.md
```diff
@@ -95,17 +95,34 @@ console.log(result.reason) // Evaluation reason
...
+if (context !== undefined) {
+  await OpenFeature.setContext(context)
+}
+
+const provider = new DatadogProvider({
+  clientToken: 'pub_...',
+  site: 'datadoghq.com',
+  env: 'production',
+  initialFlagsConfiguration: configuration,
+})
```
sameerank, line 135: I see that `applicationId` is optional in the public type

https://github.com/DataDog/openfeature-js-client/blob/afe31a76ad70ea8fbca4a7ee4c5810cc3638388e/packages/browser/src/domain/configuration.ts#L15

But `DatadogProvider` currently rejects its runtime configuration without it

https://github.com/DataDog/openfeature-js-client/blob/afe31a76ad70ea8fbca4a7ee4c5810cc3638388e/packages/browser/src/domain/configuration.ts#L92-L95

Should we include `applicationId` in this example to match the current runtime behavior?

sameerank (approved):

packages/browser/README.md
btthomas: Thank you!
https://github.com/DataDog/openfeature-js-client/pull/353/commits/d0d2f51ca49071c2ff8717ced6347ec08f757891

packages/browser/README.md
btthomas: Another PR coming soon for the optional/required bit.

Either way, fixed the README here: https://github.com/DataDog/openfeature-js-client/pull/353/commits/d0d2f51ca49071c2ff8717ced6347ec08f757891

btthomas: https://github.com/DataDog/openfeature-js-client/pull/355

sameerank:
#344 has been rebased onto the latest `main` and force-pushed at `939da976`. ~Could you rebase `blake.thomas/FFL-2837/precomputed-context` onto the updated `sameerank/FFL-2835/configuration-from-string`?~ (Never mind, I went ahead with the rebase) #346’s documented initialization flow depends on #353.

leoromanovsky (approved):
👍

PR #354, opened 2026-08-06
feat(browser): include RUM user in evaluation context

leoromanovsky:
## Motivation

Customers often set identity attributes such as email and company through the RUM SDK's `setUser()` API, but the browser OpenFeature provider currently requires those same attributes to be configured again through `OpenFeature.setContext()`. As a result, exposure events omit identity data customers already supplied to RUM, which makes it difficult to connect experiment assignments with workflows such as support-ticket analysis.

## Changes

The browser provider now enriches OpenFeature evaluation context from `DD_RUM.getUser()` when RUM integration is enabled. The RUM user ID becomes the OpenFeature targeting key, flat primitive user properties become evaluation attributes, and explicitly supplied OpenFeature fields retain precedence. The same enriched context is used for precomputed-assignment requests and exposure construction so flag assignment and the emitted exposure describe the same subject.

The browser documentation now explains initialization order, precedence, update behavior, and the flat-property constraint. Unit and end-to-end tests cover context merging, precompute request contents, exposure payloads, unavailable RUM state, and the RUM integration opt-out.

## Decisions

The existing `enableRumFeatureFlagTracking` setting remains the integration boundary, so customers who disable RUM integration do not have RUM user data copied into evaluation or exposure context. This avoids introducing a second overlapping RUM toggle.

Only string, number, and boolean RUM properties are copied. Exposure ingestion accepts flat subject attributes, so filtering nested structures in the SDK prevents a nested RUM property from invalidating the entire exposure event. Explicit OpenFeature context wins over RUM-derived values to preserve application intent.

Context enrichment happens during provider reconciliation and through the OpenFeature before hook. Exposure construction also enriches directly because context returned by one hook's before stage is not exposed to a different hook's after callback by the Web SDK.

## Verification

- `yarn workspace @datadog/openfeature-browser test --runInBand` — 156 tests passed
- `yarn workspace @datadog/openfeature-browser typecheck`
- `yarn workspace @datadog/openfeature-browser build`
- `yarn lint`
- `yarn format:check`

## Checklist

- [x] Updated documentation
- [x] Tested locally
- [ ] Tested on staging
- [x] Added unit and end-to-end test coverage

### DIFF
--- a/packages/browser/src/openfeature/exposures.ts
+++ b/packages/browser/src/openfeature/exposures.ts
@@ -5,11 +5,16 @@ import { timeStampNow } from '@datadog/js-core/time'
 import type { EvaluationDetails, FlagValue, Hook, HookContext } from '@openfeature/web-sdk'
 import type { FlaggingConfiguration } from '../domain/configuration'
 import { startExposuresBatch } from '../transport/startExposuresBatch'
+import { enrichEvaluationContextWithRumUser } from './rumIntegration'
 
 /**
  * Create hook for exposure logging.
  */
-export function createExposureLoggingHook(configuration: FlaggingConfiguration, exposureCache: AssignmentCache): Hook {
+export function createExposureLoggingHook(
+  configuration: FlaggingConfiguration,
+  exposureCache: AssignmentCache,
+  includeRumUserContext: boolean
+): Hook {
   const pageMayExitObservable = createPageMayExitObservable(configuration)
   const exposuresBatch = startExposuresBatch(
     configuration,
@@ -22,7 +27,10 @@ export function createExposureLoggingHook(configuration: FlaggingConfiguration, 
   return {
     after: (hookContext: HookContext, details: EvaluationDetails<FlagValue>) => {
       const timestamp = timeStampNow()
-      const exposureEvent = createExposureEvent(hookContext.context, details)
+      const evaluationContext = includeRumUserContext
+        ? enrichEvaluationContextWithRumUser(hookContext.context)
+        : hookContext.context
+      const exposureEvent = createExposureEvent(evaluationContext, details)
       if (!exposureEvent) {
         return
       }
--- a/packages/browser/src/openfeature/exposures.ts
+++ b/packages/browser/src/openfeature/exposures.ts
@@ -2,10 +2,9 @@ import type { Context, RawError } from '@datadog/browser-core'
 import { addTelemetryDebug, createPageMayExitObservable } from '@datadog/browser-core'
 import { type AssignmentCache, createExposureEvent, type ExposureEventWithTimestamp } from '@datadog/flagging-core'
 import { timeStampNow } from '@datadog/js-core/time'
-import type { EvaluationDetails, FlagValue, Hook, HookContext } from '@openfeature/web-sdk'
+import type { EvaluationContext, EvaluationDetails, FlagValue, Hook, HookContext } from '@openfeature/web-sdk'
 import type { FlaggingConfiguration } from '../domain/configuration'
 import { startExposuresBatch } from '../transport/startExposuresBatch'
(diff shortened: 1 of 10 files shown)


packages/browser/src/openfeature/provider.ts
```diff
@@ -98,8 +99,8 @@ export class DatadogProvider implements Provider {
     this.hooks = []
     this.events = new OpenFeatureEventEmitter()
 
-    const isRumFeatureFlagTrackingEnabled = options.enableRumFeatureFlagTracking ?? true
-    if (isRumFeatureFlagTrackingEnabled) {
+    this.isRumIntegrationEnabled = options.enableRumFeatureFlagTracking ?? true
+    if (this.isRumIntegrationEnabled) {
```
leoromanovsky, line 114: what does this accomplish?

packages/browser/src/openfeature/rumIntegration.ts
```diff
@@ -1,13 +1,45 @@
...
+import { type Context, getGlobalObject } from '@datadog/browser-core'
+import type { EvaluationContext, EvaluationDetails, FlagValue, Hook, HookContext } from '@openfeature/web-sdk'
 
 export interface DDRum {
   // biome-ignore lint/suspicious/noExplicitAny: DD RUM interface
   addFeatureFlagEvaluation: (flagKey: string, value: any) => void
+  getUser?: () => Context
+}
+
+export function enrichEvaluationContextWithRumUser(context: EvaluationContext): EvaluationContext {
```
leoromanovsky, line 10: keys that the user provides to the evaluation context have higher priority than what's in setUser, if there is a collision

leoromanovsky:
## Local validation

GitHub Actions is currently unavailable, so I ran the full repository validation locally against commit `a99daf635c51f1e1b653ce743b467de132132bec` after initializing the pinned `ffe-system-test-data` submodule (`44092ce8e5448795591fc37ef4441f077b9d34e7`).

| Command | Result |
| --- | --- |
| `yarn test` | ✅ 22 suites passed; 463 tests passed |
| `yarn typecheck` | ✅ All 3 packages passed |
| `yarn lint` | ✅ 137 files checked; no errors |
| `yarn format:check` | ✅ All files match Prettier formatting |
| `yarn build` | ✅ Core, browser, and node-server builds passed |

Test breakdown:

- `@datadog/flagging-core`: 7 suites, 275 tests
- `@datadog/openfeature-browser`: 13 suites, 156 tests
- `@datadog/openfeature-node-server`: 2 suites, 32 tests

packages/browser/src/openfeature/exposures.ts
```diff
@@ -22,7 +26,8 @@ export function createExposureLoggingHook(configuration: FlaggingConfiguration,
   return {
     after: (hookContext: HookContext, details: EvaluationDetails<FlagValue>) => {
       const timestamp = timeStampNow()
-      const exposureEvent = createExposureEvent(hookContext.context, details)
+      const evaluationContext = getEvaluationContext(hookContext.context)
```
dd-oleksii, line 29: minor: have you considered using a before-hook? We can return new context from a before-hook and it gets merged into the main context for evaluation and all subsequent hooks. This may be cleaner than piping context manually

dd-oleksii (approved):

packages/browser/src/openfeature/exposures.ts
leoromanovsky: Yea, Web SDK still types before hooks as returning void.

The Web SDK treats before as returning nothing and discards the returned object.

https://github.com/open-feature/js-sdk/blob/web-sdk-v1.10.0/packages/web/src/hooks/hook.ts#L1-L3

I'll add a todo here to migrate when supported is added.

PR #356, opened 2026-08-07
chore: remove @openfeature/server-sdk as a runtime dependency

greghuels:
## What

At a high-level, this PR removes `@openfeature/server-sdk` as a runtime dependency.

- **Custom EventEmitter** — replaces `OpenFeatureEventEmitter` from `@openfeature/server-sdk` with a custom `NodeProviderEventEmitter` built on Node's native `EventEmitter`. Zero `@openfeature` imports in compiled `.js`.
- **Bundled type declarations** — `dts-bundle-generator` inlines all `@openfeature/core` and `@openfeature/server-sdk` types into a single self-contained `index.d.ts` with zero `@openfeature` references. Necessary for SSI where `dd-trace`'s `node_modules` can't reach the customer's `@openfeature/server-sdk`.

## Why

`dd-trace-js` plans to vendor the OpenFeature provider to work around `@datadog/openfeature-node-server` having `@openfeature/server-sdk` as a peer dependency, which fails under SSI ([dd-trace-js#9570](https://github.com/DataDog/dd-trace-js/pull/9570)). This PR removes the `@openfeature/server-sdk` runtime dependency so `dd-trace-js` can instead declare `@datadog/openfeature-node-server` as a regular dependency.

This approach has several benefits:
- Reduces build complexity in dd-trace-js
- Allows for patch fixes to automatically reach customers, assuming dd-trace-js specifies this library as a patch-level range in its package.json.

### Why `build:types` is needed

`tsc` alone produces `.d.ts` files that preserve external imports:

```typescript
// cjs/provider.d.ts (tsc output)
import type { Provider, ResolutionDetails, ... } from "@openfeature/server-sdk";
import type { ServerProviderEvents, ... } from "@openfeature/core";
```

Under SSI, `dd-trace` is installed outside the application's `node_modules`, so these `.d.ts` files live in `dd-trace`'s `node_modules` tree. TypeScript resolves the `@openfeature/server-sdk` import by walking up from the `.d.ts` file's location — it never reaches the customer's `node_modules` where their `@openfeature/server-sdk` is actually installed:

```
/opt/datadog/dd-trace-js/                          ← dd-trace lives here
  node_modules/
    @datadog/openfeature-node-server/
      cjs/provider.d.ts                             ← .d.ts is here, can't resolve @openfeature/server-sdk

/home/user/my-app/                                  ← customer's app
  node_modules/
    @openfeature/server-sdk/                        ← customer's SDK is here, unreachable
```

`build:types` uses `dts-bundle-generator` to inline all `@openfeature/*` types into a single self-contained `index.d.ts` with zero external `@openfeature` imports, so the types resolve regardless of where `dd-trace` is installed.

See https://github.com/DataDog/dd-trace-js/pull/9570 for the SSI context.

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,7 +47,10 @@
   },
   "packageManager": "yarn@4.10.3",
   "resolutions": {
-    "typescript": "5.9.3"
+    "tar@npm:7.5.11": "npm:^7.5.22",
+    "minimatch@npm:10.1.1": "npm:^10.2.6",
+    "js-yaml@npm:4.1.1": "npm:^4.3.1",
+    "typescript@npm:>=5.0.2": "npm:5.9.3"
   },
   "dependencies": {
     "emoji-name-map": "^2.0.3"
--- a/package.json
+++ b/package.json
@@ -62,10 +62,5 @@
       "jest>@jest/core>jest-resolve>unrs-resolver": true,
       "lerna>nx": true
     }
-  },
-  "resolutions": {
-    "tar@npm:7.5.11": "npm:^7.5.22",
-    "minimatch@npm:10.1.1": "npm:^10.2.6",
-    "js-yaml@npm:4.1.1": "npm:^4.3.1"
   }
 }
--- a/package.json
+++ b/package.json
@@ -46,6 +46,9 @@
     "webpack": "5.106.2"
   },
   "packageManager": "yarn@4.10.3",
+  "resolutions": {
+    "typescript": "5.9.3"
+  },
   "dependencies": {
     "emoji-name-map": "^2.0.3"
   },
--- a/packages/node-server/index.d.ts
+++ b/packages/node-server/index.d.ts
@@ -1,0 +1,641 @@
+import { Channel } from 'node:diagnostics_channel';
+
+export type Metadata = Record<string, string>;
+/**
+ * Defines where the library is intended to be run.
+ */
+export type Paradigm = "server" | "client";
+export type PrimitiveValue = null | boolean | string | number;
+export type JsonObject = {
+	[key: string]: JsonValue;
+};
+export type JsonArray = JsonValue[];
+/**
+ * Represents a JSON node value.
+ */
+export type JsonValue = PrimitiveValue | JsonObject | JsonArray;
+export type EvaluationContextValue = PrimitiveValue | Date | {
+	[key: string]: EvaluationContextValue;
+} | EvaluationContextValue[];
+/**
+ * A container for arbitrary contextual data that can be used as a basis for dynamic evaluation
+ */
+export type EvaluationContext = {
+	/**
+	 * A string uniquely identifying the subject (end-user, or client service) of a flag evaluation.
+	 * Providers may require this field for fractional flag evaluation, rules, or overrides targeting specific users.
+	 * Such providers may behave unpredictably if a targeting key is not specified at flag resolution.
+	 */
+	targetingKey?: string;
+} & Record<string, EvaluationContextValue>;
+export type FlagValueType = "boolean" | "string" | "number" | "object";
+/**
+ * Represents a JSON node value.
+ */
+export type FlagValue = boolean | string | number | JsonValue;
+export type ResolutionReason = keyof typeof StandardResolutionReasons | (string & Record<never, never>);
+/**
+ * A structure which supports definition of arbitrary properties, with keys of type string, and values of type boolean, string, or number.
+ *
+ * This structure is populated by a provider for use by an Application Author (via the Evaluation API) or an Application Integrator (via hooks).
+ */
+export type FlagMetadata = Record<string, string | number | boolean>;
+export type ResolutionDetails<U> = {
+	value: U;
+	variant?: string;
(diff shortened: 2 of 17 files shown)


package.json
```diff
@@ -46,6 +46,12 @@
     "webpack": "5.106.2"
   },
   "packageManager": "yarn@4.10.3",
+  "resolutions": {
+    "tar@npm:7.5.11": "npm:^7.5.22",
+    "minimatch@npm:10.1.1": "npm:^10.2.6",
+    "js-yaml@npm:4.1.1": "npm:^4.3.1",
```
greghuels, line 52: Pre-existing security fixes that would have been overwritten in the yarn.lock. This codifies the install so that `yarn.lock` regeneration doesn't affect manually upgraded dependencies.

package.json
```diff
@@ -46,6 +46,12 @@
...
+    "typescript@npm:>=5.0.2": "npm:5.9.3"
```
greghuels, line 53: dts-bundle-generator's typescript >=5.0.2 dependency caused yarn to resolve TS 7.0.2

packages/node-server/package.json
```diff
@@ -25,27 +25,27 @@
...
   "scripts": {
-    "prepack": "yarn build",
+    "prepack": "yarn build && node scripts/verify-no-openfeature-dep.js",
     "pack": "yarn pack",
-    "build": "run-p build:cjs build:esm",
+    "build": "run-p build:cjs build:esm && yarn build:types",
     "build:cjs": "rm -rf cjs && tsc -p tsconfig.cjs.json && yarn replace-build-env cjs",
     "build:esm": "rm -rf esm && tsc -p tsconfig.esm.json && yarn replace-build-env esm",
+    "//build:types": "Bundles @openfeature/* types into a self-contained index.d.ts with no external imports, required for SSI where dd-trace's node_modules can't reach the customer's @openfeature/server-sdk. See https://github.com/DataDog/dd-trace-js/pull/9570",
+    "build:types": "trap 'rm -f src/.dts-entry.ts' EXIT; printf '/// <reference path=\"./global.d.ts\" />\\nexport * from \"./index\"\\n' > src/.dts-entry.ts && dts-bundle-generator -o index.d.ts src/.dts-entry.ts --external-inlines @openfeature/core @openfeature/server-sdk --no-banner --export-referenced-types false",
```
greghuels, line 34: `tsc` produces `.d.ts` files that preserve external imports (`import type { ... } from "@openfeature/server-sdk"`). Under SSI, `dd-trace` is installed outside the application's `node_modules`, so TypeScript can't resolve those imports — the `.d.ts` files live in `dd-trace`'s `node_modules` tree and can't reach the customer's `@openfeature/server-sdk`.

`dts-bundle-generator` inlines all `@openfeature/*` types into a single self-contained `index.d.ts` with zero external imports, so the types resolve regardless of where `dd-trace` is installed.

See https://github.com/DataDog/dd-trace-js/pull/9570 for the SSI context.

packages/node-server/src/provider-event-emitter.ts
```diff
@@ -0,0 +1,83 @@
...
+  EventHandler,
+  Logger,
+  ProviderEventEmitter,
+  ServerProviderEvents,
+} from '@openfeature/core'
+
+/**
+ * OpenFeature-compatible event emitter that does not require an OpenFeature SDK at runtime.
+ */
+export class NodeProviderEventEmitter implements ProviderEventEmitter<ServerProviderEvents> {
```
leoromanovsky, line 14: Please add tests for add/remove/remove-all, duplicate handlers, async rejection, etc.

packages/node-server/index.d.ts
```diff
@@ -0,0 +1,598 @@
+import { Channel } from 'node:diagnostics_channel';
```
leoromanovsky, line 1: Please add a test to exercise type-checks and runs against minimum and current supported OpenFeature SDK versions.

leoromanovsky (commented):
lgtm! Please increase test coverage in the support version range.

packages/node-server/src/provider-event-emitter.ts
greghuels: Yeah, good call. This is implicitly tested with integration tests, but the class itself could use some unit tests.

packages/node-server/index.d.ts
greghuels: I'll update in https://github.com/DataDog/openfeature-js-client/pull/357, since I already have typecheck tests there as well.

leoromanovsky: Ok I see now, that's fine.

leoromanovsky (approved):
Oops meant to pre-approved pending the two comments.

packages/node-server/src/provider-event-emitter.ts
greghuels: updated!

packages/node-server/index.d.ts
greghuels: Merged into this PR and now running against the min supported openfeature version

PR #364, opened 2026-08-12
Guard precomputed ETag reuse by context

leoromanovsky:
## Motivation

`fetchPrecomputedConfiguration` accepts a caller-supplied previous configuration. When that configuration belongs to context A but the new request targets context B, the public fetcher currently sends context A’s ETag and can reuse A’s assignments after a `304 Not Modified` response.

This PR is stacked on #351 to demonstrate and close that context-boundary gap.

## Changes

- validate `previousConfiguration` against the requested context inside `fetchPrecomputedConfiguration`
- use only the validated configuration when setting `If-None-Match` or handling `304`
- add a regression test that models context A returning `304` when its ETag is incorrectly sent, while context B returns fresh assignments when it is omitted

## Decisions

The invariant lives in the exported public fetcher rather than only in `createFlagsConfigurationFetcher`, so direct callers receive the same context-safety guarantee as the provider wrapper. Rules configurations remain unchanged because they are not precomputed for an evaluation context.

Validation:

- `yarn workspace @datadog/openfeature-browser test --runInBand test/transport/fetchConfiguration.spec.ts` — 25 tests passed
- `yarn workspace @datadog/openfeature-browser typecheck`
- `yarn biome check packages/browser/src/transport/fetchConfiguration.ts packages/browser/test/transport/fetchConfiguration.spec.ts --diagnostic-level=error`


packages/browser/src/transport/fetchConfiguration.ts
```diff
@@ -89,12 +89,15 @@ export async function fetchPrecomputedConfiguration(
   options: PrecomputedConfigurationFetchOptions
 ): Promise<FlagsConfiguration> {
   const url = buildConfigurationUrl(options, 'precomputed')
+  const previousConfiguration = configMatchesContext(options.previousConfiguration, options.context)
+    ? options.previousConfiguration
+    : undefined
```
dd-oleksii, line 94: major: reusing etag across context must be fine. If context change results in a different response, that response must have a different etag. If context changed in a way that does not affect the response, it should get the same etag (e.g. change in attribute that is not used for evaluation).

So having some etag is better than none. We also want to avoid busting cache on insignificant attribute change.

dd-oleksii (changes requested):

PR #365, opened 2026-08-13
feat: support semantic version conditions

greghuels:
## Summary

- Add SEMVER evaluation to `@datadog/flagging-core`.
- Support `SEMVER_EQ`, `SEMVER_NEQ`, `SEMVER_LT`, `SEMVER_LTE`, `SEMVER_GT`, and `SEMVER_GTE`.
- Run all shared FFE evaluation cases through `DatadogNodeServerProvider`.
- Add focused core SemVer unit coverage.
- Pin `ffe-system-test-data` to `aaa97e4`, which has the SEMVER system tests.

## Why uint64-safe string comparison

The evaluator keeps SemVer core components as decimal strings instead of converting them to JavaScript `Number` values.

- **Compatibility:** The FFE SemVer behavior is shared across SDKs. The Go/Rust-compatible implementation accepts `uint64`-bounded core components, including values above `Number.MAX_SAFE_INTEGER`, so narrowing JavaScript to `Number.MAX_SAFE_INTEGER` would create cross-SDK differences.
- **Performance:** Length-plus-digit string comparison avoids `Number` parsing on every evaluation and is faster than converting the component strings to numbers at comparison time. It also avoids precision loss for large valid versions.

### DIFF
--- a/packages/core/src/evaluation/evaluateForSubject.ts
+++ b/packages/core/src/evaluation/evaluateForSubject.ts
@@ -4,7 +4,7 @@ import { type TimeStamp, timeStampNow } from '../time'
 import { TargetingKeyMissingError } from './errors'
 import { createEvaluationTimestampMetadata } from './evaluationMetadata'
 import { matchesShard } from './matchesShard'
-import { isValidRule, matchesRule, type Rule } from './rules'
+import { hasInvalidSemverComparand, isValidRule, matchesRule, type Rule } from './rules'
 import { type Flag, type Split, type VariantType, variantTypeToFlagValueType } from './ufc-v1'
 
 export function evaluateForSubject<T extends FlagValueType>(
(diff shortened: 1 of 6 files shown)


packages/core/src/evaluation/evaluateForSubject.ts
```diff
@@ -157,15 +157,25 @@ function validateTypeMatch(expectedType: FlagValueType, variantType: VariantType
...
+  }
+
+  if (
+    flag.allocations.some(
       (allocation) =>
-        Array.isArray(allocation.splits) &&
-        allocation.splits.every((split) => Array.isArray(split.shards)) &&
-        (allocation.rules === undefined ||
-          (Array.isArray(allocation.rules) && allocation.rules.every((rule) => isValidRule(rule))))
+        Array.isArray(allocation.rules) && allocation.rules.some((rule) => hasInvalidSemverComparand(rule))
```
dd-oleksii, line 167: minor: shall this be moved to `isValidRule`?

packages/core/src/evaluation/semver.ts
```diff
@@ -0,0 +1,231 @@
...
+    return { value: '0', next: start + 1 }
+  }
+
+  let end = start
+  while (end < version.length && isAsciiDigit(version.charCodeAt(end))) {
+    end++
+  }
+
+  const value = version.slice(start, end)
+  if (value.length > MAX_UINT64.length || (value.length === MAX_UINT64.length && value > MAX_UINT64)) {
```
dd-oleksii, line 105: minor: why do we artificially limit the max value?

packages/core/src/evaluation/semver.ts
```diff
@@ -0,0 +1,231 @@
...
+  patch: string
+  prerelease: string
+}
+
+/**
+ * Parse the SemVer subset.
+ * Core identifiers are limited to uint64; numeric prerelease identifiers may
+ * be arbitrarily large.
+ */
+export function parseSemver(version: unknown): ParsedSemver | null {
```
dd-oleksii, line 20: nitpick: the current implementation seems to be verbose and likely slow'ish

I have a feeling that a regex could be faster and shorter. Another option is to split on `.` and `-` and validate components afterwards

dd-oleksii (approved):

packages/core/src/evaluation/semver.ts
greghuels: For parity with implementation in other dd-trace libraries. It shouldn't matter too much either way, since this won't be a real-world use case.

packages/core/src/evaluation/semver.ts
greghuels: Screenshot 2026-08-13 at 9 11 00 AM

I had an LLM run through some benchmarks on the 2 implementations you suggested along with the current implementation. It looks like the current implementation is still the fastest.

PR #366, opened 2026-08-13
test(browser): execute packed package in Chromium

leoromanovsky:
## Motivation

PR #344 adds production browser entry points and a React Native package smoke test, but the existing browser package check stops after TypeScript and Vite compilation. A package can build successfully while failing when its production output is loaded in a real browser, especially around package exports, initialization order, runtime globals, and capability-specific entry points.

This PR is stacked on #344 to close that runtime coverage gap before those browser APIs merge.

## Changes

- add a Playwright Chromium smoke test to the isolated Vite test app
- build and serve separate pages for the full and `/precomputed` browser entry points from packed core/browser tarballs
- exercise protobuf configuration parsing, round-trip serialization, boolean and integer evaluation, and SHA-256 matching through the full entry point
- rerun the full entry point with `TextEncoder`, `TextDecoder`, and `BigInt` unavailable
- verify the precomputed entry point parses and round-trips precomputed data without retaining rules
- fail on browser page exceptions and `console.error` output
- update the package-install check to install only the Chromium headless shell, execute the smoke tests, and restore the root lockfile after temporary package wiring

## Decisions

- keep this stacked on `sameerank/FFL-2835/configuration-from-string` because it directly tests the APIs introduced by #344
- test packed artifacts through a production Vite build and preview server rather than workspace links or a DOM emulator
- keep the existing module-level assertion as the proof that `/precomputed` excludes Protobuf-ES; this browser smoke complements it by proving the entry point executes correctly
- install only Chromium's headless shell to limit CI download size while retaining the real browser runtime boundary

## Validation

- `yarn build`
- `yarn lint`
- `yarn format:check`
- `yarn test`
- `yarn test:browser-install` (3 Playwright tests passed)

### DIFF
--- a/test-app/src/precomputed.ts
+++ b/test-app/src/precomputed.ts
@@ -1,0 +1,39 @@
+import { configurationFromString, configurationToString } from '@datadog/openfeature-browser/precomputed'
+import { assert, reportSuccess } from './smoke'
+
+const response = {
+  data: {
+    attributes: {
+      createdAt: '2026-07-06T23:01:56.822Z',
+      flags: {
+        'precomputed-flag': {
+          allocationKey: 'allocation',
+          variationKey: 'on',
+          variationType: 'BOOLEAN',
+          variationValue: true,
+          reason: 'STATIC',
+          doLog: true,
+        },
+      },
+    },
+  },
+}
+const configuration = configurationFromString(
+  JSON.stringify({
+    version: 1,
+    precomputed: { response: JSON.stringify(response) },
+    rules: { response: 'ignored by the precomputed entrypoint' },
+  })
+)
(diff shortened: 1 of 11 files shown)


test-app/src/precomputed.ts
```diff
@@ -0,0 +1,39 @@
...
+    },
+  },
+}
+const configuration = configurationFromString(
+  JSON.stringify({
+    version: 1,
+    precomputed: { response: JSON.stringify(response) },
+    rules: { response: 'ignored by the precomputed entrypoint' },
+  })
+)
```
leoromanovsky, line 27: ok good but what about proto test?

vjfridge (approved):
The new coverage lgtm, thanks!

sameerank (approved):
Agreed with adding more test coverage 🙌

